PR #1, opened 2019-06-26
smirnoff99Frost parameter coverage set

ChayaSt:
This PR adds the script to generate input conformers and `cmiles` identifiers for an `OptimizationDataset`.


j-wags:
Thanks for pulling this together, @ChayaSt. I think this is a good process to start using.

Speaking of process, I think it'd be best to make a standard practice be `conda list > requirements.txt`. Chaya is an expert with cheminformatics toolkits and molecule data set handling, but future submitters may not be, and more provenance is always a good thing. What do you think, @ChayaSt , @dgasmith and @yudongqiu?

ChayaSt:
@j-wags, I agree. That is better practice. I updated `requirements.txt`.

dgasmith:
LGTM, @yudongqiu can you build out a fractal submission script?

yudongqiu:
@dgasmith Sure no problem. Do we want to merge #3 first?

PR #3, opened 2019-06-28
Use updated input smiles

ChayaSt:
@btjanaka and @davidlmobley updated the input SMILES file here https://github.com/openforcefield/open-forcefield-data/pull/29 so this PR uses the updated input file to generate QCArchive input molecules. 

It also loads the molecule that causes oequacpac to segfault separately so we can get input conformations for it.


dgasmith:
@yudongqiu lets merge this first, though you can probably work concurrently.

Once we have submitted our first batch, we will need to think some on how to do future batches which are more of updates to previous ones.

yudongqiu:
@dgasmith Sounds good. The input and output files are updated so I think it's better to merge this first. BTW, I don't have write access to merge here. 😬

dgasmith:
You should have access pending now.

PR #4, opened 2019-06-28
add create_optimization_dataset.py

yudongqiu:
@dgasmith I'm not exactly sure if this is what you expect, but I added a script that converts the input molecules and cmiles_ids into arguments of the `client.add_procedure()` function, such as:
```
("optimization", "geometric", options, initial_molecule)
```
If there's an example to build the `OptimizationDataset`, I can try to convert this to a script that directly builds an `OptimizationDataset` from the JSON file, but I don't have a way to test it yet.


dgasmith:
Can you migrate this into building a [OptimizationDataset](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/tests/test_collections.py#L483)? Index is the canonical smiles index, is this right @ChayaSt?

ChayaSt:
@dgasmith, for the index, do you mean the label of the job? 
Since these are unrestrained optimization, the labels should just be `cmiles_identifiers['canonical_isomeric_smiles']`.

yudongqiu:
@ChayaSt Each molecule have one unique `cmiles_identifiers`, but may have multiple comformers. If we run one geometry optimization for each conformer, that means some of the some of them will have duplicate `cmiles_identifiers['canonical_isomeric_smiles']`. Is that fine?

yudongqiu:
I guess we want to continue to use `{index}-1`, `{index}-2` to indicate conformers for the same molecule?

ChayaSt:
Yes, that is a solution.

yudongqiu:
@dgasmith I just rewrote the script. Since I'm not finally submitting it, could you please try that and see if it works?
Two things that might need to be updated:
1. The client configure file is not included here
2. The conda environment showing in requirements.txt might not be the same between @ChayaSt who runs `generate.py` and @dgasmith who will run `create_optimization_dataset.py`

yudongqiu:
@dgasmith I just tested the creation script on my local server and fixed an issue. It should be good now.
Screen Shot 2019-06-28 at 11 27 35 PM

6-25-19-smirnoff99Frost-coverage/create_optimization_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    ds: qcportal.collections.OptimizationDataset
+        The dataset created
+
+    """
+    client = ptl.FractalClient.from_file(client_config_file)
+    # create a new dataset with specified name
+    print(f"Creating OptimizationDataset < {dataset_name} >")
+    ds = ptl.collections.OptimizationDataset(dataset_name, client=client)
+    # create specification for this dataset
+    opt_spec = {"program": "geometric"}
```
dgasmith: Do we need to specify other geometric options here for this dataset?

6-25-19-smirnoff99Frost-coverage/create_optimization_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    qc_spec = {"driver": "gradient", "method": qm_method, "basis": qm_basis, "program": "psi4"}
```
dgasmith: Do we need bond-orders or similar computations for this dataset?

.gitignore
```diff
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# vscode
+.vscode/
```
dgasmith: As a note, standard practice is to add this files to your own personal `~/.git/info/exclude`. I don't particularly mind, but you will get pushback in other OSS projects.

6-25-19-smirnoff99Frost-coverage/create_optimization_dataset.py
yudongqiu: I think the bond orders are not used in the fitting, but if @ChayaSt want that information we can add that.

6-25-19-smirnoff99Frost-coverage/create_optimization_dataset.py
yudongqiu: I think the default options should be good for this free optimization with no constraints.

.gitignore
yudongqiu: I see. Thanks for the advice!

PR #7, opened 2019-06-28
Grid Optimization Dataset for Trivalent Nitrogens Database

jmaat:
This PR contains the following: 
(a) Jupyter notebook which contains a script with an example job submission for a 1-d improper grid optimization
(b) Approximately 300 .sdf files of molecules for grid optimization calculations


j-wags:
For this submission, it might be good to 
1) Nest all of the new files (notebook and data) in a subdirectory, like `06-28-19-Nitrogen-grid-optimization`
2) Put all the SD files in a single folder and zip it up (GitHub is bad at showing hundreds of files)
3) run `conda list > requirements.txt` in the new folder as well, to record the exact state of your environment when the script was run

yudongqiu:
I agree with @j-wags, and I feel that jupyter-notebooks might be harder for tracking changes and reproducing the submit. It would be better if we convert it to a python script. 
BTW, is this dataset planned to be submitted to the public QCFractal server, or it's intended to run on a local server maintained by @jmaat?

jmaat:
@yudongqiu I would like this can be submitted to the public QCFractal server.

j-wags:
@yudongqiu -- I get what you mean about the notebooks making it hard to track changes, but I'd prefer to make it less of a burden on the submitter, so I'd be fine if it stays as a notebook. Though, @jmaat, at the same time it would be great if we removed any non-functional code in the notebook, so that it's clearer what the process was when we read through it later.

@dgasmith -- The current code assembles the jobs for direct submission. Is there an easy way to convert them to a JSON file for you?

j-wags:
> BTW, is this dataset planned to be submitted to the public QCFractal server, or it's intended to run on a local server

@yudongqiu Submitted to the public server, I think

dgasmith:
Typically we write them out as a dictionary of `key: job_description` JSON pairs so that I know what to call each optimization. As a note there is no need to add `optimization_spec` and `qc_spec` to each JSON entry as those are static across all files. After you build the `service` object, you can use `json.loads(service.json())` to get a JSON-style python dictionary that we can be written via Python. I forgot a helper method there which does this for you.

One thing that occurred to me is do we know that if freezing the coordinate at +/- 40 degrees is accessible? As in there are no steric clashes with will cause quantum chemistry failures? If not, we might need to add an optional similar to torsion drives that stops the computation after accessing energy states that are say ~50 kcal/mol above the lowest.

yudongqiu:
@j-wags I agree the notebook can serve as a submission script when it's cleaned up. One thing I have in mind is that when we later assemble the pieces in these scripts into a pipeline, it would be much easier if we can re-use the functions/classes defined in these scripts directly. When converting a notebook to a python script, I'm imagining that people would wrap up what they did in the notebook kernel and refactor that for better reproducibility.

I feel that we have the similar idea when you say "removed any non-functional code in the notebook". I just have the feeling that it takes very little effort from there to convert the notebook to a well-structured python script at this time, rather than trying to understand what's going on several months later.

jmaat:
Quick update, I am planning on holding off on submission of these calculations for the next couple weeks. I'll turn the notebook into a script, and add a README.

jmaat:
Hi @dgasmith, I updated the PR. 

The main changes in this PR are the following:
-README added
-Molecules_to_run folder is zipped
-Python script for the send_mols_to_server.py added (converted from jupyter notebook).

For the README, I denoted the collection as "GridOptimization". I'm not sure if this is the correct collection name so this may need to be changed.

Please let me know if you need any additional changes.

j-wags:
@jmaat -- It looks like the `Molecules_to_run/` folder didn't get fully deleted. Could you double check that the `git rm` got pushed to Github?

j-wags:
Looks great. Thanks @jmaat!

dgasmith:
@j-wags @jmaat I closed this too early. As I do not have OpenEye I need the raw JSON data as produced by `send_qm_job`. Effectively, run this script, but instead of submitting tasks, build a large JSON dictionary and drop the data to a file. In addition, I need the index CMILES string of the indices in question like we have from @ChayaSt torsion drives.

PR #10, opened 2019-07-02
Tweaks Coverage submit script

dgasmith:
Several small tweaks to the coverage submission script. Primarily this hard-codes several parameters that were previously present on the command line. This is done to make it much more explicit *exactly* what is going into these datasets.


j-wags:
👍 

Great. _Excatness_ is _exactly_ what we want here!

...

I'll show myself out.

yudongqiu:
Sure I was trying to put the submission methods into a function that can be later reused, but if we don't want such flexibility and exactness is preferred then we can go with a flat script.

yudongqiu:
It's great to see the requirements.txt is updated cleanly.

Issue #11, opened 2019-07-02
Potential dataset: VEHICLe


j-wags:
(Edited by @jchodera):

To ensure good scaffold coverage, it was suggested we include the [VEHICLe dataset](ftp://ftp.ebi.ac.uk/pub/databases/chembl/VEHICLe/) (Virtual Exploratory heterocyclic library) from [Heteroaromatic rings of the future](http://doi.org/10.1021/jm801513z). This dataset contains ~25K heteroaromatic ring systems, some of which existing current drugs, and many of which are likely synthesizable.

For future reference, the compressed dataset is here: [VEHICLe.csv.gz](https://github.com/openforcefield/qca-dataset-submission/files/3351946/VEHICLe.csv.gz)

Reference:

William R. Pitt, David M. Parry, Benjamin G. Perry, and Colin R. Groom.
Heteroaromatic rings of the future.
J Med Chem 52:2952, 2009.
[[DOI]](http://doi.org/10.1021/jm801513z) [[FTP]](http://ftp.ebi.ac.uk/pub/databases/chembl/VEHICLe/) [[PDF]](https://github.com/openforcefield/qca-dataset-submission/files/3351948/jm801513z.pdf)

jchodera:
I'm working on constructing this dataset right now.

dgasmith:
This dataset is currently being computed 10k/25k tasks complete.

dgasmith:
This set is largely computed. Closing.

PR #12, opened 2019-07-02
Add VEHICLe OptimizationDataset

jchodera:
To ensure good scaffold coverage, it was suggested we include the [VEHICLe dataset](ftp://ftp.ebi.ac.uk/pub/databases/chembl/VEHICLe/) (Virtual Exploratory heterocyclic library) from the paper ["Heteroaromatic rings of the future"](http://doi.org/10.1021/jm801513z). This dataset contains ~25K heteroaromatic ring systems, some of which existing current drugs, and many of which are likely synthesizable. After expanding protonation and tautomeric states, the dataset contains 84761 molecules.

For future reference, the compressed dataset is here: [VEHICLe.csv.gz](https://github.com/openforcefield/qca-dataset-submission/files/3351946/VEHICLe.csv.gz)

Reference:

William R. Pitt, David M. Parry, Benjamin G. Perry, and Colin R. Groom.
Heteroaromatic rings of the future.
J Med Chem 52:2952, 2009.
[[DOI]](http://doi.org/10.1021/jm801513z) [[FTP]](http://ftp.ebi.ac.uk/pub/databases/chembl/VEHICLe/) [[PDF]](https://github.com/openforcefield/qca-dataset-submission/files/3351948/jm801513z.pdf)

Addresses #11


jchodera:
Here are some examples of the VEHICLe set:
image

dgasmith:
What priority are these compared to the grid optimizations and Hessian computations? We can always toss this in under "low" priority and let them run opportunistically.

jchodera:
From @ChayaSt :
> Checking the `generate.py` script you used, tautomers were set to False so you only expanded protonation states and stereoisomers. I set tautomers to False by default because I find that it generates a bunch or resonance structures.

I'm trying to figure out if it is OK to set `tautomers=True` or if this would cause major problems.

ChayaSt:
> I'm trying to figure out if it is OK to set tautomers=True or if this would cause major problems.

If you set it to True you will get many resonance structures that aren't necessarily different tautomer. Ideally we should filter out the resonance structures. To do that programmatically we need to check if the chemical graphs are different if all bond orders are the same. 
This is something we should add to `fragmenter` after expanding tautomers. 
See https://github.com/openforcefield/fragmenter/issues/6

jchodera:
From @ChayaSt :
> I see in the PDF you posted that there are some duplicate molecules. There's an open issue for that as well here: https://github.com/openforcefield/fragmenter/issues/33

jchodera:
@ChayaSt  : When I just enumerate stereochemistry (and not protonation states or tautomers), it looks like we *always* generate duplicates:
```
12:02:48 INFO fragmenter: Enumerating stereoisomers for thiazino[5,4-e]thiadiazine-4,5-dione
12:02:48 INFO fragmenter: 2 states were generated for c1c2c(c(=O)nns2)c(=O)sn1
INFO:root:Found duplicate canonical SMILES c1c2c(c(=O)nns2)c(=O)sn1
```
Is this potentially a bug with `fragmenter`'s stereochemistry enumeration? Should it always be generating two copies even when there *are* no stereocenters?

jchodera:
I've updated the dataset, omitting protonation states (since these looked very odd). The dataset now includes only 24867 entries.

jchodera:
The PDF has been updated too:
image

ChayaSt:
> Is this potentially a bug with fragmenter's stereochemistry enumeration? Should it always be generating two copies even when there are no stereocenters?

This might be an issue with `OEFlipper` that I need to look into. However, if the SMILES are the same it can easily be eliminated.

jchodera:
> Can you add duplicates.smi to the manifest? 

Great catch! I've added this.

> Also, if skipped_ions.smie and cmiles_failures.smi are empty, it might be better to remove them.

I'll keep them for now, since the fact that they're empty means nothing was skipped/failed!

jchodera:
This should be ready to submit to QCFractal once we have sufficient available resources.

dgasmith:
If we are happy with the input, we can go ahead and submit them.

PR #22, opened 2019-07-05
VEHICLe Dataset Submission

dgasmith:
Several small tweaks to the submission script to make the creation of very large datasets easier. This is something that will be patched up in Fractal and eventually deprecated.


dgasmith:
Forcing a merge so that I can add the Hessian computations.

PR #25, opened 2019-07-07
Fragmenter stability benchmark

ChayaSt:
This adds the 4 Idelalisib fragments to the stability benchmark.

The other fragments are currently on hold because most of those torsion involved some atoms that were in rings and had difficulty converging. This series is a bit less restricted. 

The series of Idelalisib fragments in this set are [here](https://github.com/choderalab/fragmenter_data/blob/master/frag_opt/compare_growth_path/Idelalisib_bond_17_29_to_run.pdf):


6-3-19-Torsion-Stability-Benchmark/Idelalisib_frags.json
```diff
@@ -0,0 +1,110 @@
...
+      "inchi_key": "GSVFENCKYURWTI-UHFFFAOYSA-N",
+      "molecular_formula": "C7H9N5",
+      "standard_inchi": "InChI=1S/C7H9N5/c1-2-8-6-5-7(10-3-9-5)12-4-11-6/h3-4H,2H2,1H3,(H2,8,9,10,11,12)",
+      "unique_protomer_representation": "CCN=c1c2c([nH]cn1)nc[nH]2",
+      "unique_tautomer_representation": "CCN=c1[nH]cnc2[nH]cnc12"
+    },
+    "provenance": {
+      "canonicalization": "cmiles_v0.1.5_openeye_2019.Apr.2",
+      "central_rot_bond": [
+        "[17, 29]"
```
dgasmith: Should this be a list?

ChayaSt: Yes, It should be. Will fix it.

ChayaSt:
Conflicts are resolved.

ChayaSt:
> I guess this is an improvement to the record of a previous submission?

Since the submission changed so much (many of the calculations were put on hold) so I updated the `generate.py` script to reflect that. So this script only has the input for the scans we are still running.

PR #26, opened 2019-07-08
add 2019-07-01-smirnoff99Frost-coverage-torsion

yudongqiu:
This PR adds the creation script for submitting torsiondrive scans for the coverage set.

I picked a name of the dataset "SMIRNOFF Coverage Torsion Set". Please update if another name is preferred.

@dgasmith After submitting this to my local server, I got an error message from the server when it tries to start the scan:
```
[E 190708 02:30:02 web:1788] Uncaught exception POST /service_queue (127.0.0.1)
    HTTPServerRequest(protocol='https', host='localhost:7777', method='POST', uri='/service_queue', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/home/qyd/codes/anaconda3/envs/qcf/lib/python3.6/site-packages/tornado/web.py", line 1697, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/home/qyd/codes/anaconda3/envs/qcf/lib/python3.6/site-packages/qcfractal/queue/handlers.py", line 89, in post
        self.storage, self.logger, service_input, tag=body.meta.tag, priority=body.meta.priority))
      File "/home/qyd/codes/anaconda3/envs/qcf/lib/python3.6/site-packages/qcfractal/services/services.py", line 48, in initialize_service
        storage_socket, logger, service_input, tag=tag, priority=priority)
      File "/home/qyd/codes/anaconda3/envs/qcf/lib/python3.6/site-packages/qcfractal/services/torsiondrive_service.py", line 60, in initialize_from_api
        initial_molecule=[x.id for x in service_input.initial_molecule],
      File "/home/qyd/codes/anaconda3/envs/qcf/lib/python3.6/site-packages/qcfractal/services/torsiondrive_service.py", line 60, in <listcomp>
        initial_molecule=[x.id for x in service_input.initial_molecule],
    AttributeError: 'NoneType' object has no attribute 'id'
[E 190708 02:30:02 web:2246] 500 POST /service_queue (127.0.0.1) 8.34ms
```

Could you please help me to check the source of this error?


dgasmith:
It looks like it could not find one of the molecule in the initial_molecule set. Are you adding molecules via id's or full geometries?

yudongqiu:
> It looks like it could not find one of the molecule in the initial_molecule set. Are you adding molecules via id's or full geometries?

I tried to use a list of `qcelemental.models.Molecule` objects in ds.add_entry(), and I also tried to use the list of JSON dicts, both gives me the same error.

If I use the list of id's, the client directly pops out an error:
```
Traceback (most recent call last):
  File "03_create_torsiondrive_dataset.py", line 65, in <module>
    ds.add_entry(canonical_torsion_index, initial_molecules, torsion_atom_indices, grid_spacings, energy_upper_limit=0.05, attributes=attributes)
  File "/home/qyd/codes/anaconda3/envs/openff/lib/python3.7/site-packages/qcportal/collections/torsiondrive_dataset.py", line 100, in add_entry
    molecule_ids = self.client.add_molecules(initial_molecules)
  File "/home/qyd/codes/anaconda3/envs/openff/lib/python3.7/site-packages/qcportal/client.py", line 360, in add_molecules
    return self._automodel_request("molecule", "post", {"meta": {}, "data": mol_list}, full_return=full_return)
  File "/home/qyd/codes/anaconda3/envs/openff/lib/python3.7/site-packages/qcportal/client.py", line 213, in _automodel_request
    raise TypeError(str(exc))
TypeError: 1 validation error
data -> 0
  value is not a valid dict (type=type_error.dict)
```

yudongqiu:
@dgasmith BTW, this error
```
AttributeError: 'NoneType' object has no attribute 'id'
```
only occurred to only one of the jobs when server starts to run a list of 20 jobs I think.

dgasmith:
Can you give me the schema of the molecule(s) that are giving you issues in a gist?

yudongqiu:
@dgasmith There are a total of 3849 molecules in the "initial_molecules" list of 585 torsion scans. I'm not exactly sure which one is causing the problem. It may be too big for a gist. Can I send you a json file instead?
[all_molecules.json.zip](https://github.com/openforcefield/qca-dataset-submission/files/3374976/all_molecules.json.zip)

dgasmith:
When you run `ds.add_entry` it will fail on a single entry. Which one is that and what are the molecules involved? This will narrow things down considerably.

yudongqiu:
@dgasmith I think it doesn't. The server only prints out the 

> When you run `ds.add_entry` it will fail on a single entry. Which one is that and what are the molecules involved? This will narrow things down considerably.

Sorry it doesn't. The error only occurs on the server after `ds.compute()` is called.

dgasmith:
Interesting. I will have to check it out next week.

The other thing to try is to call `client.add_molecules([])` for each `initial_molecule` list and check the returned id list to see if it is ever shorter than the input molecule list or if it contains `None`s.

yudongqiu:
@dgasmith Good idea! I will try that.

yudongqiu:
@dgasmith When I add the molecules, all of them gives valid IDs back.

I hacked the part in the server code and got 4 specific molecules that are causing this error in the TorsionDrive Dataset.

Could you please take a look? https://gist.github.com/yudongqiu/c0ed06e62474d7f52ea6c5e9d96cb11d

yudongqiu:
One thing I noticed when looking at the IDs returned by `add_molecules`, is that sometimes they are duplicated. This means the conformers generated by `generate.py` also have duplicates. I will try to remove the duplicates from the `initial_molecules` list and try again.

yudongqiu:
@dgasmith Actually duplicated molecules in the `initial_molecules` list might be the problem. After removing duplicates, the error does not happen on server anymore.

In the future I think it will be great if the QCFractal client can detect duplicates and raise an error, or if the QCFractal server can remove duplicates automatically.

This dataset is ready to go now.

dgasmith:
Ah, that does make sense how the code is structured. What we should do is change the QCFractal type from a `List[ObjectId]` to `Set[ObjectId]` so that it will automatically prevent duplicates in the input.

This line [here](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/collections/torsiondrive_dataset.py#L19).

yudongqiu:
@dgasmith Sounds like a nice way to improve this. Shall we proceed to submit this?

dgasmith:
Sounds good, I am away at the moment. Would mind making the change to Fractal as well?

dgasmith:
Do we need bond orders here? I can get this in tonight or tomorrow.

yudongqiu:
@dgasmith Sure if that's not too expensive, we can compute and save bond orders for each final molecule for future use. Do you know what's the best way to do that?

yudongqiu:
@dgasmith I tried that your method does fixed the error. I have created a PR with that fix: https://github.com/MolSSI/QCFractal/pull/320

PR #34, opened 2019-07-25
Biphenyls

ChayaSt:
This is the initial biphenyl set used as motivation for fragmenter.


ChayaSt:
@dgasmith, these are ready to go.

dgasmith:
Hessians are nearly done, pulling this in.

PR #35, opened 2019-07-26
Create dataset_priority

j-wags:
Adds a file containing a discrete ranking of dataset priority.

@dgasmith Let me know if this is sufficiently detailed (or modify this branch directly to correct).


j-wags:
@ChayaSt I assume that your dataset is the Biphenyl set in the PR. Please correct that if I'm wrong.

@jchodera, please add other datasets in any order that you see fit. This is an effort to make sure we prioritize the release-1 critical sets first, but if you're comfortable ranking the others, go ahead.

PR #36, opened 2019-07-28
Coverage Hessian Submission

### LINKED ISSUE #33: submit hessian jobs for the coverage set of molecules
The optimizations for the coverage set of molecules were submitted in
https://github.com/openforcefield/qca-dataset-submission/tree/master/2019-06-25-smirnoff99Frost-coverage

We want to submit hessian calculation for the optimized geometries. 
The simplest way is to submit one hessian computation for each optimization record. The downside is that it is possible there are duplications in the optimized geometries, and we will most likely not use all of the data in fitting.

Since we are more interested in the hessian of the global minimum of each molecule, we can pick the lowest energy structure for each molecule out of the optimized geometries, and just submit hessian jobs for those. This will greatly reduce the total cost of this dataset.

dgasmith:
Submits coverage set, closes #33.

### DIFF
--- a/2019-06-25-smirnoff99Frost-coverage/04_create_hessian_dataset.py
+++ b/2019-06-25-smirnoff99Frost-coverage/04_create_hessian_dataset.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+
+import copy
+import json
+import tqdm
+import tarfile
+from collections import Counter
+#import qcportal as ptl
+import qcfractal.interface as ptl
+from qcelemental.models import Molecule
+
+# Updates the dataset rather than recomputing
+UPDATE = False
+#UPDATE = True
+
+
+print("Initializing dataset...")
+client = ptl.FractalClient.from_file()
+#client = ptl.FractalClient("localhost:7777", verify=False)
+
+# create or pull a previous dataset with specified name
+
+# Pull the optimization dataset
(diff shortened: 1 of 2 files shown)


yudongqiu:
Thanks for making this submission quick. My understanding is that this is a temporary solution for the Hessian calculations, because it creates a general `Dataset`, that does not have "entry" and "attributes". In the future we might want to refactor this when `HessianDataset` is available, which supports `add_entry()`, `gen_entry()` and the entries should have attributes such as `cmiles_ids`.

BTW, I think this PR also contains changes in the file `2019-07-05 eMolecules force field discrepancies 1/02_create_optimization_dataset.py`, is that intended?

dgasmith:
I don't think we will make a HessianDataset as there are no changes required to the code to support energy/gradient/Hessian/property single computations within the Dataset class. We can certainly expand the Dataset class to work similarly to the ProcedureDataset classes like Optimization/TorsionDrive/etc.

Yes, the update to eMolecules was intended here.

yudongqiu:
Sure sounds good. I think if the `Dataset` class is expand to support the `add_entry()` and `get_entry()` methods, that would be enough for the hessian jobs.

davidlmobley:
Is this ready to go now?

yudongqiu:
> Is this ready to go now?

I believe this is already submitted and running on the server.

dgasmith:
Just need an approval and merge.

davidlmobley:
OK, merging!

PR #37, opened 2019-07-30
SMIRNOFF Coverage Torsion Set Submission

dgasmith:
Submission creation.


yudongqiu:
I have a question here, I think this PR is for submitting torsiondrives for the `

However, the title says "coverage torsion set", and in the script 
```
dataset_name = "SMIRNOFF Coverage Torsion Set"
```

Is this a mistake? Will it overwrite or mix with the existing "SMIRNOFF Coverage Torsion Set" that I submitted here? https://github.com/openforcefield/qca-dataset-submission/blob/master/2019-07-01-smirnoff99Frost-coverage-torsion/03_create_torsiondrive_dataset.py

yudongqiu:
I just queried the server and found that we have a "SMIRNOFF Coverage Torsion Set" and a "SMIRNOFF Coverage Torsion Set 1" on the server.

From the number of jobs in each dataset, I guess  "SMIRNOFF Coverage Torsion Set" contains data for the phenyl set, and "SMIRNOFF Coverage Torsion Set 1" contains data for the actual coverage torsion set.

dgasmith:
Whoops, mistake on my part. It is very hard to overwrite a collection, so unlikely to ever accidentally replace one.

What is the correct name for this so I can change it?

davidlmobley:
Urgh, that's not ideal. Probably would have been better to call this something like "SMIRNOFF Phenyl Torsion Set 1" or something.

davidlmobley:
@yudongqiu ?

yudongqiu:
I'm not the owner of this dataset, I think @ChayaSt is the submitter from the README.

PR #41, opened 2019-08-16
Adding .json file for job submission of NitGridOpt molecules

jmaat:
I'm adding the file `nitrogen_Jobs.json` to the Nitrogen Grid Optimization folder. The dictionary in the .json file contains the file name as keys and the json for job submission for respective file name key. 

@dgasmith Please let me know if you need anything else.


jmaat:
Hi @dgasmith, this is follow-up on this PR. Can we submit the jobs in the added .json? Do you need anything additional?

dgasmith:
Looking through this now, a few notes:
- I think you serialized the symbols/geometries before going into JSON so they are represented as strings. We can convert, but might be worth looking at your pipeline so that this doesn't happen in the future.
- I think we are missing the CMILES identifiers. Do we want to go ahead and run without?
- Key names are currently "./Molecules_to_run/eMolecules_1000_562.sdf" for our main molecule index what do we want to call each of these molecules, "1000_562"?
- You have keywords as `5d5731c7796432a5d45474ec` for the quantum computations. What keywords did you want here? I can come up with reasonable values, mostly did you want to store bond order and other one-electron properties for these computations?

jmaat:
-I updated the key names to just the ID number of the molecule (for example:  "./Molecules_to_run/eMolecules_1000_562.sdf" > "1000_562"). Can I call each of the molecules by the unique key in the dictionary?
-No CSMILES for now is okay and sorry about the strings for the symbols and geometries. I had converted them to strings because I got a type error: `TypeError: Object of type ndarray is not JSON serializable` when I did not convert. Can we still submit the jobs? I can fix this for the future.
-In regards to the keywords, I want geometries, optimization energies, and WBOs (specifically for the three bonds to the nitrogen). What keyword should I replace the current one with? Can I use the same keyword for all the jobs?

davidlmobley:
@jmaat I could see your data being also broadly useful within OpenFF (e.g. perhaps we'll want to fit to some of it) so I think that probably means we want CMILES, since all of our datasets have these... unless I'm missing something?

jmaat:
@davidlmobley Okay, I agree I want to use CSMILES more immediately especially if we want to start using this more broadly for fitting. I'll make updates to the PR.

jmaat:
@davidlmobley I added CMILES and CMILES ids to the .json. @dgasmith Can you advise on the correct quantum computation keyword to store WBOs, geometries, and optimization energies? Is it necessary to include these? I notice some of the other .json files in this repository do not include any keywords.

dgasmith:
I can add this when I pull it in.

dgasmith:
Let me try a few of these with a RDKit backend tomorrow. If everything works out I will merge it in!

jmaat:
@dgasmith Sounds great!

jmaat:
@dgasmith Any updates? I am hoping to get this submitted soon.

PR #45, opened 2019-09-04
geometry optimization input for Roche molecules

ChayaSt:
This PR adds the script to generate conformers for geometry optimization of Roche molecules. 

**Note**:
This script was written before we had a standard way of submitting to QCArchive and it also used a much older version of `fragmenter` that still had the `workflow_api.py` module. It is **not** recommended to generate optimization inputs like this.


yudongqiu:
Thanks a lot! This will serve as an important record for the Release-1 fitting data generation. I suggest that we merge this soon.

davidlmobley:
Just to be clear: This was or was not used in fitting?

@yudongqiu if you've reviewed you can go ahead and merge... unless you don't have permissions, in which case LMK and I can (a) give you permissions, and (b) merge.

yudongqiu:
@davidlmobley Yes was used in fitting. Sure I will merge this.

PR #50, opened 2019-09-07
[WIP] Add Pfizer "100 challenging fragments" dataset to prep for pipeline

davidlmobley:
Adds Pfizer 100 fragments dataset. This should be prepped as a torsion drive dataset, presumably, as these were selected as fragments where QM and force field torsional profiles differed substantially. No fragmentation should be needed.

To do:
- [ ] Prep for dataset submission


jchodera:
Looks like I had already set this up locally as an optimization dataset but hadn't merged it. I'll see if I can merge both in.

davidlmobley:
The important thing is to get it in/ready -- both optimizations AND torsions would be great. 

I'm not surprised you'd already started, as you said on Slack you would, but then it never showed up so I was not sure the status.

jchodera:
Superceded by #55 and #56, so closing this.

PR #66, opened 2019-10-24
Connectivity added to nitrogen grid optimization dataset 

jmaat:
Three changes were made:

1. connectivity was added to the .json
2. in send_mols_to_server.py the script was adjusted to add connectivity to the initial molecule
3. the range of angles in steps was adjusted to go from -50 to 50 (prior was -40 to 40)


2019-06-28-Nitrogen-grid-optimization/send_mols_to_server.py
```diff
@@ -71,7 +77,7 @@ def send_qm_job(ptl_mol, nitrogen, nitrogen_i,  mol):
                     "scans": [{
                         "type": "dihedral",
                         "indices": list(indices),
-                        "steps": [-40, -36, -32, -28, -24, -20, -16, -12, -8, -4, 0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40],
+                        "steps": [-50, -48, -46, -44, -42,-40, -36, -32, -28, -24, -20, -16, -12, -8, -4, 0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 42, 44, 46, 48, 50],
```
dgasmith: Is there a reason for a finer grid at the edges here? As a note we do not have technology which can stop a gridoptimizaiton if the energy becomes too large.

@yudongqiu In torsiondrive for 2d-scans how do you stop the propagation if the energy is too high? In the 1-D case it is easy to see the cap. For 2-D we have something like where `x` is complete and `z` is too high. We can view a propagation like so:

```
o o o o o
o o x z o
o o o o o 
```

```
o o x o o
o x x z o
o o x o o 
```

```
o x x x o
x x x z o
o x x x o 
```

We can quickly see that in the 2-D case we can overcome bad area even though we might not want to?

jmaat:
Yes the reasoning for that is after going through the 1-d scans I noticed a lot of the energy profiles had their minimas partially cut off. I wanted to extend the angle range a little to hopefully capture that. Here's an example of what I mean (referring to the blue QM trend line). 
[['i5']_C1C=CS[N@]1Br.pdf](https://github.com/openforcefield/qca-dataset-submission/files/3773093/i5._C1C.CS.N%40.1Br.pdf)

Would be great if there's a way to stop the propagation if the energy gets too high!

dgasmith:
Curious here in why the change in the step size (2 vs 4 degrees).

jmaat:
I changed it and there's no reason for that. Good catch

PR #73, opened 2019-12-05
[WIP] Add 2019-12-05-OpenFF-Benchmark-Primary-1-torsion

hyejang:
This is created for OpenFF Benchmark Primary set 1 torsiondrive submission. 

259 1-D torsions were selected from the benchmark primary set 1 and its torsion parameter coverage is 66/156. No molecular size filter has been used.


dgasmith:
As discussed, waiting for the bulk of the current HF-3c computations from @ChayaSt to filter through before submitting.

PR #74, opened 2019-12-11
2-D Nitrogen Grid Optimization Dataset

jmaat:
This PR contains a new directory that has the 2-d grid optimization for a set of diverse trivalent nitrogen containing molecules. The 2-d grid optimizations scan the valence and improper angle simultaneously. 

The new directory has a README.md which gives an description of the different files in the directory.


dgasmith:
Thanks, as per the new torsiondrives lets hold off on submission for a few days until we patch up the docker images and fix the hf-3c issues.

PR #78, opened 2020-01-17
Add 2020-01-17-OpenFF-Benchmark-Full-1-torsion

hyejang:
This is created for OpenFF Benchmark Full set 1 torsiondrive submission.

595 1-D torsions were selected from the benchmark full set 1 and its torsion parameter coverage is 139 / 156.


dgasmith:
Have we already computed these in different sets?

hyejang:
We have `OpenFF Full Optimization Benchmark 1` in `OptimizationDataset` collection.

dgasmith:
I added this as `OpenFF Primary Benchmark 2 Torsion Set` since we already have a `1`.

PR #89, opened 2020-03-25
Hessians for optimization data set

jmaat:
@dgasmith These are hessians from the recently submitted gen 2 optimization data sets.


2020-03-20-OpenFF-Gen-2-Hessian-Set-1-Roche/03_create_hessian_dataset.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+
+import copy
```
dgasmith: We should add these to the pre-existing folders I think as they are still steps in the same dataset. If we continue with the separate dataset strategy we should not duplicate the first steps.

jmaat: Ok, removed the additional directories and added `03_create_hessian_dataset.py` to preexisting optimization directories. @dgasmith

dgasmith:
It looks like the Hessian directors are still in the PR.

jmaat:
Hessian directories should be removed now, @dgasmith

PR #90, opened 2020-03-27
Add 2nd round Bayer torsion set and supplemental set

hyejang:
2 torsiondrive datasets were added. Total number of 1-D torsions is 241.

1. `OpenFF Gen 2 Torsion Set 5 Bayer 2` : 219 1-D torsions (17 torsion scans are expected to be reused)
2. `OpenFF Gen 2 Torsion Set 6 supplemental 2`: 22 1-D torsions (3 torsion scans are expected to be reused)


hyejang:
For Bayer set, since it contains large molecules ( molecules w/ more than 30 heavy atoms), I filtered out the large molecules before selecting torsions. One concern was the possibility of losing chemical space of the molecule set after excluding large molecules. But the similarity matrix shows that we are not losing chemical diversity from the filtering. 

image

hyejang:
@dgasmith  could you check the two sets i added? They look good to go.

PR #99, opened 2020-05-13
fixed wrong dihedral indices

ChayaSt:
Several dihedral indices were wrong because of a bug in the generation script. This PR fixes that bug and the dihedral indices in the input JSON.


dotsdl:
This has been submitted; new dataset name is `"OpenFF Fragmenter Validation 1.0"`.

PR #103, opened 2020-06-05
OpenFF DANCE 2 eMolecules t142 Random v1.0

btjanaka:
This job is based on the one in `2020-06-01-DANCE-1-eMolecules-t142-selected`, but with a different dataset consisting of random molecules with the t142 parameter. I also fixed some mistakes in the README in that directory. I also added visualizations to it.


trevorgokey:
Just as a pedantic note, visualizing first dataset was informative; a PDF visualizing these drives would also be great as a reference and debugging tool in the case of any future issues.

btjanaka:
Sounds good.

trevorgokey:
I will leave some comments here regarding the DANCE 1 set as that PR has already been merged. The comments should generally apply to this set as well.

1. Torsions on a ring need to be filtered out
2. **The torsion indices are incorrect.** They should correspond to the raw indices of the QCMolecule.
3. There are 20 torsiondrives, but 2071 input molecules, with one in particular containing 305 inputs. I would try to prune this space down as much as possible. Torsion drives will report as an error if even one of these inputs fail.

btjanaka:
@jmaat

btjanaka:
I think to solve 2), we would have to generate the indices based on the `canonical_smiles` (see https://github.com/openforcefield/qca-dataset-submission/pull/103/files#diff-fa297b772d74b09c62b0ae282337726fR122-R124; there's also identical lines in the DANCE 1 directory)

dotsdl:
Holding off on submission until @trevorgokey's points have been addressed.

For point (3) we would *either* like to:
- trim down the conformers to a smaller set of "meaningful" conformers for each molecule.
    - e.g. choose conformers that won't be arrived at by the torsion drive itself (not easy)
- clearly state that for molecules with many conformers, we expect many failures, but aren't sure which conformers will fail; the strategy is to try many and take the ones that succeed.

dotsdl:
@jmaat is this a submission you are taking ownership of? We have some questions that we'd like addressed before we can proceed with submission. Thanks!

PR #102, opened 2020-06-04
Add Torsion Dataset for molecules with t142 parameter from DANCE

btjanaka:
Hi, I am an undergraduate working with @jmaat in the Mobley Lab. We are trying to run torsion drive calculations on a dataset of 20 molecules we generated with the [DANCE pipeline](https://dance.rtfd.io), a project we are currently working on. This is my first time submitting a PR to this repo; would you mind taking a look at my code and letting me know if I have formatted the job correctly? I have left some comments on lines where I was uncertain about certain details. Thanks!

PR #116, opened 2020-07-24
OpenFF DANCE 1 eMolecules t142 v1.0 - tracking

dotsdl:
This PR plugs submission #102 into our [lifeycle](https://github.com/openforcefield/qca-dataset-submission/blob/master/LIFECYCLE.md) automation.

- [ ] Ready to submit!

### DIFF
--- a/submissions/2020-06-01-DANCE-1-eMolecules-t142-selected/.gitignore
+++ b/submissions/2020-06-01-DANCE-1-eMolecules-t142-selected/.gitignore
@@ -1,9 +1,0 @@
-# Editor files
-.*.swp
-
-# Output files -- only `t142_selected.smi` (from DANCE) and
-# `optimization_inputs.json.gz` should be included.
-*.smi
-*.sdf
-*.txt
-*.json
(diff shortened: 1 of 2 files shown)


dotsdl:
@jthorton or @trevorgokey, this is meant to be a [retcon](https://en.wikipedia.org/wiki/Retroactive_continuity) in order to plug the dataset into our automated lifecycle, starting with error cycling. Unfortunately, since it predates our use of QCSubmit, it means our validations won't pass, and we may not be able to meet all our [STANDARDS](https://github.com/openforcefield/qca-dataset-submission/blob/master/STANDARDS.md).

For datasets like these, I'm happy to prepare the PRs as I've done here, but I'd like a second set of eyes to review and merge. Basically, we just want to make sure we think the dataset will be able to be processed by our error cycling, which should only mean `dataset.json` must be a changed file in the PR, and it must have:
* `dataset_name`
* `dataset_type`

Can you think of anything else we should do for these kinds of datasets?

trevorgokey:
Sounds good @dotsdl. 

Yes, I think in the future we should ensure, for OpenFF datasets, that we have the `default` spec as a name, and that name refers to `b3lyp-d3bj/dzvp`. We might just do this automatically if the PR doesn't include this spec. For example if the `dataset.json` has some other spec in it, we can just add our `default` in. Cool thing about QCA is that if `b3lyp-d3bj/dzvp` is a spec included as a different name, creating a copy of the spec with the name `default` will not produce extra tasks.

I'll update the STANDARDS to reflect this.

PR #124, opened 2020-08-04
Disaccharides: added two example .json files for the QCArchive queue

dscerutti:
- [ ] Ready to submit!


dotsdl:
Awesome, thank you for submitting this @dscerutti!

I will start a notebook for pulling this into a form for submission; I'll engage @jthorton as necessary with questions.

jthorton:
@dscerutti do you have the original input file you used in gmx for this? we might be able to use that to generate the cmiles attributes for the molecule which are currently missing. Also, do you know how many unique molecules there are and roughly how many conformers for each molecule?

dotsdl:
Some notes from my first pass on this submission:
1. @dscerutti please adjust your JSON writer to not include a comma for the last element in a sequence.
2. @dscerutti there is a comma on line 239 in each file (e.g. [here](https://github.com/openforcefield/qca-dataset-submission/pull/124/files#diff-dea8056a0d4a09d6c854ba066fc43906R239)). Please adjust your JSON writer to leave out the comma.
3. @dscerutti can you verify that these molecules have no net charge?
4. @dscerutti to echo @jthorton: what format are the structure files you start with? We may be able to use those instead of SDFs for extracting CMILES.

dscerutti:
I have fixed the issues mentioned in my .json writer.  Not hard.  How shall I post the new outputs?

dotsdl:
Hey @dscerutti, feel free to push new versions of the JSON files we've got here. You can copy them over the existing ones in your `DS_Contribution` branch, commit, and push. Thanks!

dotsdl:
I also added a [Jupyter notebook](https://github.com/openforcefield/qca-dataset-submission/blob/DSC_Contribution/submissions/2020-07-28-OpenFF-Disaccharide-Optimization-Set/Dataset%20Preparation.ipynb) giving the process as it stands for taking these initial files toward a submission. We want to:

1. Read in any number of molecule input files produced by @dscerutti. These may take the form of individual JSON files for each molecule, or a concatenated set of JSON structures (perhaps separated with a `---`) to avoid thousands of files in this repository.
2. Create an `OptimizationDataset` with QCSubmit, and populate it from the files in (1). This gains us many of the automatic components QCSubmit provides, such as metadata, in the `OptimizationDataset`.
3. Add all molecules from the input files (with existing, pre-optimized conformations) to our `OptimizationDataset` object. We will pass these through `openforcefield.topology.Molecule` for similar reasons as in (2). This will require us to be able to generate CMILES from the provided data, which will require some separate structure file(s).
4. Export a single `dataset.json` to be validated, then submitted.

@jthorton does this sound like a reasonable approach?

@dscerutti what is the format of the structure files you use as inputs to MDGX? Forgive my memory if you've mentioned this previously. I'm looking for an optimal route to CMILES in (3).

submissions/2020-07-28-OpenFF-Disaccharide-Optimization-Set/DAllpa1-2DAllpa1-OME.json
```diff
@@ -0,0 +1,326 @@
+{
+  "dataset_name": "mdgx-created data set",
+  "dataset_tagline": "A data set created by mdgx",
+  "dataset_type": "OptimizationDataSet",
+  "method": "d3bj b3lyp",
```
trevorgokey, line 5: This model won't be incorporated unless it is in a spec

dotsdl:
@j-wags do you have an opinion on the pathway we take for getting CMILES from @dscerutti's molecules? Passing SDF files in through the toolkit has been our answer so far, but before we fully commit to that I wanted to know if there are any (better) alternatives?

If not, then I'd like to ask @dscerutti to produce SDF files for his molecules. He'll have to make an SDF writer to do this.

j-wags:
I think SDF is the best option by far for this pipeline. Mol2 is fraught with ambiguity, and our only reader for it is non-open-source.

submissions/2020-07-28-OpenFF-Disaccharide-Optimization-Set/DAllpa1-2DAllpa1-OME.json
trevorgokey: Actually this is my mistake, I believe this is how QCSubmit handles these fields

trevorgokey:
(Carrying over from #125)
Thanks for the switch @dotsdl. 

The utility script `util/generate_cmiles_attributes.py` #125 to generate CMILES (without connectivity) is not OpenFF supported software. It is provided as a best-effort based on limited knowledge, and the approach of guessing CMILES from a 3D structure is not 100% accurate. This aspect is not a short-term issue for the QM calculations as CMILES is not a required component, but we must ensure that all molecules we submit are MM compatible, and proper CMILES information is essential for this. 

@dscerutti Until we have a "standard" approach to do this, we need to rely on you for the accuracy of the CMILES information generated. Please use the CMILES info generated with the utility as a starting point, check it for accuracy, and update the `json` with the appropriate `attributes` it generates.

Also note that the util with minor modification can generate the appropriate fields if a well-formed `sdf` file is available. If it is not present, the least-common-denomenator is a connectivity table. If we have symbols, a 3D representation, and connectivity, we should have everything we need.

Thanks for the contribution! We are trying to solidify the submission process based on past experience, so putting in the extra work in the beginning will pay dividends at the end of the pipeline.

j-wags:
Apologies for the late comment on this -- I agree with Trevor that we don't want to be guessing connectivity if it's not provided. This is the same as why we can't load molecules from PDB+CONECT. 

It's very important in our workflow that the initial graph molecules (the canonical isomeric explicit hydrogen tagged SMILES) be 100% correct, otherwise we can wind up with some mangled/mispercieved molecules going into our FF fitting. 

Especially in this case, where MDGX knows the full information about these molecules, it doesn't make sense to guess the connectivity. SDF would contain the full data content of the graph molecule before it loses it on the way to become a QCA molecule, and it's super important that that graph is recorded accurately.

dscerutti:
Does the .json format itself include a means for incorporating connectivity?  I am working with the connectivity in a different way for a related problem right now, but I could easily add that to the .json.  The SDF format, requiring some assessment of formal charges, is a ways off unless you want me to continue making my best ability to code freshman chemistry principles into assigning formal charges.

trevorgokey:
Yes, alongside `geometry`, `symbols`, and others, there is a `connectivity` field that we eventually need, so it would be great if you can supply that information. See recent submissions for the complete molecule format needed by QCSubmit, e.g. [2020-07-27-theory-bm-set-b3lyp-d3bj-dzvp/dataset.json#L51](https://github.com/openforcefield/qca-dataset-submission/blob/master/submissions/2020-07-27-theory-bm-set-b3lyp-d3bj-dzvp/dataset.json#L51). The closer we get to that molecule spec, the easier it will be to automate.

j-wags:
Yeah. The workable options, in terms of descencing goodness, are
* SDF
* Mol2 (but the kind that can be read by OpenEye)
* QCSchema (JSON) with connectivity including bond orders. 

Gotta run for a family call. I think Trevor has the same understanding as I do. Ping my on Slack and we can organize a meeting for Monday morning.

dscerutti:
Yes, I had seen that earlier but fell short of adding it.  Also the molecular charge, and multiplicity are things that I can provide.  The user must supply the multiplicity, because at my current level of knowledge that is an "ask a chemist" question, but multiplicity 1 is the default and seems to have worked for all of the neutral disaccharides I did.

trevorgokey:
Yes, we don't handle any other multiplicities other than 1 at the moment. The charge is definitely a required component of a molecule. I think we should start with the simpler spec I linked above; QCSubmit handles most of the other aspects. So, far starters, we need a json that looks something like this if SDF/mol2 are not available (we would essentially read an SDF on each molecule and then produce the following):
```
{

"mol_name_1":  {

    "initial_molecules": [ 
        { "symbols": ["H", "H"], "connectivity": [ [0, 1, 1.0 ] ], "geometry": [ 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 ] } ], "molecular_charge": 0, "molecular_multiplicity": 1 }, 
        { "symbols": ["H", "H"], "connectivity": [ [0, 1, 1.0 ] ], "geometry": [ 0.0, 0.0, 0.0, 0.0, 0.0, 1.2 ] } ], "molecular_charge": 0, "molecular_multiplicity": 1 } 
    ],

    "attributes": { CMILES info from cmiles.generator.get_molecule_ids and validated by the user }

}
```
This is the level of "truth" we assume that the user (e.g. the data curator) has provided and is correct, since we essentially ship these values straight to the QM procedure. We are working on validators and filters to help with this to prevent bad inputs, but it's a work in progress. Note that this does not have any constraints/torsiondrive indices, so this would be an input for a single point or optimization. Also please do not submit initial molecules that only differ by connectivity, as these appear as the same molecule in QCArchive at the current moment.

Once we figure out the molecules, we can move to QM specs, procedure definition, etc.

dscerutti:
Can the .json include coordinates as "geometry": [ 0.0, 0.0, 0.0, 0.0, 0.0, 1.0 ]", or even

"geometry": [
   0.0, 0.0, 0.0,
   0.0, 0.0, 1.0 ]

or am I restricted to one number per line?  If I could print the geometry and other aspects of the structure in a more compact fashion I'd love to do away with the leading white space and save screen space / disk space.

dotsdl:
@dscerutti 

Thanks for fixing the JSON issues! There remain a couple:
1. `"scf_properties"`: the list has a comma on the last element
2. `"initial_molecules"` : the list has a comma on the last element

It may be worth trying to run these through a JSON-parsing utility as a check before committing. Either of these will throw errors if there's something wrong with the JSON:
1. `python -m json.tool <file.json>`
2. `cat <file.json> | jq`

JSON generally doesn't care about indentation or newlines, but the ultimate test is whether these tools can parse it without error.

dscerutti:
Just updated the two .json files.  Thanks for showing me that validation script--I've got the mdgx output passing that filter, letting it return a .json in the long, x/y/z coordinate on separate rows format as it reads my output.  Dealt with a couple of other spurious commas to make it happen.  Oh, still need to add connectivity...

dscerutti:
Connectivity added and passing the JSON interpreter.

dscerutti:
Saw this shortly after you posted, but I was getting over a migraine.  The box where I've got all my progress on these disaccharides is down--I was probably running too many of its GPUs at once in addition to Dave Case's own calculations, which in the past has caused it to fall over.  It will be tomorrow morning before the box can be rebooted.  But, there is supposed to be metadata in the files, as there is in other GeomeTRIC inputs that mdgx can write, so probably another easy fix albeit not as simple as A=>Bohr or addition of bonded connections.

dotsdl:
Thank you both! From our call, it sounds like we technically only need constraints added to the JSON files, and @jthorton will have everything he needs to form a submission. 

Speaking further with Josh, however, we would still like one PDB file for each molecule (so one per folder in the tarball). This allows us to validate that our resulting JSON submission, however complete and correct in terms of spec, actually matches our desired molecules. This simple check can save us a lot of pain and/or confusion later.

So, we need two things in the tarball:
1. Constraints added to each JSON file.
2. A PDB file for each molecule (one per directory).

@dscerutti, do you anticipate any issues with delivering item (2) in addition to (1)?

jthorton:
@dscerutti Thanks for updating the JSON files with the constraints, unfortunately, I am unable to open the files as there is a trailing comma in the constrain atom indices list, can you please remove this and reupload the files.

jthorton:
@dscerutti I have updated the dataset to use the new files please have a look over the pdf attached. Note the toolkits have Identified every molecule as being unique that is 976 unique optimizations does this sound correct? as I though some inputs were just different conformations of the same molecules.

PR #125, opened 2020-08-07
Add utility to generate CMILES info for dataset submission when connectivity is absent

trevorgokey:
This small utility is able to generate the molecule `attributes` field, which is required by nearly everywhere dealing with QCSchema, and is required for dataset submission. Showing the CLI help below; it has a simple API to allow plugging the results into a dataset during submission.
```
usage: generate_cmiles_attributes.py [-h] [--use-rdkit] [--use-openeye]
                                     input_json

Generates CMILES data for use with QCArchive when connectivity is not known.
Requires cmiles, the openff toolkit, OpenEye for protomers, RDkit for the
other attributes, and openbabel

positional arguments:
  input_json     input file in json format. The json should be a list of
                 QCSchema-style symbols and coordinates (in bohr). A
                 dict(geometry=xyz, symbols=syms) is all that is needed for
                 each molecule.

optional arguments:
  -h, --help     show this help message and exit
  --use-rdkit    CMILES generator should use rdkit when possible
  --use-openeye  CMILES generator should use only OpenEye
```

### DIFF
--- a/util/generate_cmiles_attributes.py
+++ b/util/generate_cmiles_attributes.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 import json
 import sys
-import numpy as np
 import os
+import io
 
+import numpy as np
 
 def dict_to_xyz_string(mol):
     bohr2ang = 0.529177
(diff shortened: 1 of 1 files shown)


dotsdl:
Awesome, thank you @trevorgokey! Giving this a spin. I'll likely merge this one directly into #124 so we can use it there, and since this builds on that branch.

dotsdl:
Looking great @trevorgokey! I'm merging this onto #124 to continue there. I switched to using StringIO for reading the SDF string as a file with the toolkit. I *think* this should work, but will find out there.

util/generate_cmiles_attributes.py
```diff
@@ -29,10 +30,15 @@ def qcmiles(jsmol, toolkit='rdkit'):
     obConversion.ReadString(obmol, xyz_str)
     sdf = obConversion.WriteString(obmol)
 
+    with io.StringIO(sdf) as sdf_stream:
```
trevorgokey, line 33: Jeff mentioned this the other day! Thanks for doing this

PR #131, opened 2020-08-14
add compute for benchmark ligands.

jthorton:
This PR will add a new compute specification to the OpenFF-benchmark-ligand-fragments-v1.0 dataset to run an unconstrained version of parsley. 

The management script may need to be updated to also accept these `compute.json` files.

- [ ] Ready to submit!


jthorton:
I think we can close this as the constraints were automatically removed by `openmmforcefields`.

PR #132, opened 2020-08-14
Added individual prod environments for different QCEngine harnesses

dotsdl:
We have recently found cases where certain libraries in our conda environment installs can conflict with one another in their dependencies, especially across different channels. To mitigate this in our production QCArchive worker environments, we have decided to create separate environments for different task types, namely:
- MM (OpenMM, openforcefield, ambertools, etc.)
- QM (psi4, dftd3, gcp, etc.)
- ML (torchani, pytorch)

This PR adds these production environments. It also updates the existing `qcarchive-worker-openff.yaml` environment currently used for QM+MM deployments, which currently covers common workload types we are computing. We have decided to pin our MM dependencies to specific versions so we can deliberately choose when to upgrade these in production.

### DIFF
--- a/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
+++ b/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
@@ -7,10 +7,10 @@ channels:
 dependencies:
   - python =3.7
   - qcfractal
-  - qcengine =0.15.0
+  - qcengine =0.16.0
 
   # ML calculations
-  - pytorch
+  - pytorch =1.6.0
 
   # procedures
   - geometric
@@ -19,4 +19,4 @@ dependencies:
   - dask-jobqueue
 
   - pip:
-    - torchani
+    - torchani =2.2
--- a/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
+++ b/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
@@ -1,11 +1,11 @@
 name: qcarchive-worker-openff-ani
-channel_priority: flexible
 channels:
   - conda-forge
-  - pytorch-nightly
+  - pytorch
   - defaults
 dependencies:
   - python =3.7
+  - pip
   - qcfractal
   - qcengine =0.16.0
 
@@ -19,4 +19,4 @@ dependencies:
   - dask-jobqueue
 
   - pip:
-    - torchani =2.2
+    - torchani ==2.2
--- a/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
+++ b/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
@@ -1,0 +1,22 @@
+name: qcarchive-worker-openff-ani
+channel_priority: flexible
+channels:
+  - conda-forge
+  - pytorch-nightly
+  - defaults
+dependencies:
+  - python =3.7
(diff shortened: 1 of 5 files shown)


dotsdl:
@trevorgokey or @jthorton, would one of you mind reviewing to make sure I captured what we need for these deployment types? Thanks!

trevorgokey:
I don't see why these wouldn't work. One thing we may want to do is pin all important packages (e.g. torchani) since the version provenance degrades a bit in the pipeline, and QCA deduplication is not sensitive to these package versions. Having a commit history here with explicit versions can potentially help with debugging as well and make sure all openff-qcf workers are perfectly in-sync.

devtools/prod-envs/qcarchive-worker-openff-ani.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - pytorch
+
+  # procedures
+  - geometric
+
+  # compute backends
+  - dask-jobqueue
+
+  - pip:
+    - torchani
```
dotsdl, line 22: Let's pin this version.

devtools/prod-envs/qcarchive-worker-openff-ani.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - conda-forge
+  - pytorch-nightly
+  - defaults
+dependencies:
+  - python =3.7
+  - qcfractal
+  - qcengine =0.15.0
+
+  # ML calculations
+  - pytorch
```
dotsdl, line 13: Let's pin this version.

dotsdl:
@trevorgokey agreed, thanks for pointing this out. I've added comments to the spots we want to pin before we merge this. I've solicited version preferences from @jchodera if he has them.

devtools/prod-envs/qcarchive-worker-openff-ani.yaml
jthorton: this should be at least 2.1.1 to get the ANI2x model as well.

devtools/prod-envs/qcarchive-worker-openff-ani.yaml
```diff
@@ -0,0 +1,22 @@
+name: qcarchive-worker-openff-ani
+channel_priority: flexible
+channels:
+  - conda-forge
+  - pytorch-nightly
+  - defaults
+dependencies:
+  - python =3.7
+  - qcfractal
+  - qcengine =0.15.0
```
jthorton, line 10: This needs to be pinned to the next release version to support ANI2x.

dotsdl: Done!

devtools/prod-envs/qcarchive-worker-openff-ani.yaml
dotsdl: Done!

dotsdl:
I think we should now be good to go! Pinned to latest QCEngine (0.16.0) on all envs, and pinned the ANI deployment to pytorch 1.6.0 and torchani 2.2 (both latest, and recommended from everyone we've got feedback from).

dotsdl:
This should be good to go. Thanks @trevorgokey and @jthorton for the testing and feedback!

PR #134, opened 2020-08-18
Torsiondrive restart fix.

jthorton:
It looks like the management script was only keeping track of the errored torsiondirves for the last specification it checks this is now changed to capture all records making sure all torsiondrives are restarted across all specs.


dotsdl:
Ah, I spaced and didn't make the same change to the torsiondrive error handling as we did for the optimizations to handle this issue. Thanks for identifying and fixing this @jthorton!

PR #136, opened 2020-08-21
Add ligand spec

jthorton:
This will add the ANI2x computations to the fragmented ligand benchmark set. We expect the CI to fail here on basis check as some molecules contain I and Br which are not part of the ANI2x model.

I also need to update the conda version of QCSubmit to allow users to submit jobs which raise basis set errors as I am not sure I want to allow this or if we should make a new dataset which will only contain the correct elements.

- [ ] Ready to submit!


jthorton:
@dotsdl QCSubmit is now updated and I have updated the lifecycle script to use the new `ignore_errors` flag we should be good to start these now when the lifecycle update to use the `compute.json` files is finished.

submissions/2020-07-27-OpenFF-Benchmark-Ligands/compute.json
```diff
@@ -0,0 +1,63 @@
...
+      "method": "ani2x",
+      "basis": null,
+      "program": "torchani",
+      "spec_name": "ani2x",
+      "spec_description": "ANI2x ML potential.",
+      "store_wavefunction": "none"
+    }
+  },
+  "dataset_name": "OpenFF-benchmark-ligand-fragments-v1.0",
+  "dataset_tagline": "OpenForcefield TorsionDrives.",
```
dotsdl, line 13: Does the tagline need to match that of the original dataset? Do we want it to?

submissions/2020-07-27-OpenFF-Benchmark-Ligands/compute.json
```diff
@@ -0,0 +1,63 @@
...
+  "maxiter": 200,
+  "driver": "gradient",
+  "scf_properties": [
+    "dipole",
+    "quadrupole",
+    "wiberg_lowdin_indices",
+    "mayer_indices"
+  ],
+  "priority": "normal",
+  "description": "A TorsionDrive dataset using geometric.",
```
dotsdl, line 24: Same as above question. I anticipate this doesn't matter, but thought it worth asking. I don't consider it a blocker for submission today.

dotsdl:
Since this is supplemental compute, we care less about checks passing; we know this is exploratory. Thanks @jthorton!

dotsdl:
Moved to "Requires Scientific Review" as this submission is no longer advancing. It may be possible to clear the ~200 `OptimizationRecord` errors due to lack of convergence by increasing the `maxiter` key for the `optimization_procedure` used.

Do you have thoughts on whether this would be effective @JoshHorton? @trevorgokey?

PR #137, opened 2020-08-21
Draft of the first official STANDARDS (v3)

trevorgokey:
This PR represents our first set of guidelines and standards that OpenFF datasets must conform to. Since we currently have datasets labeled with version 1 and 2, we will start the version of our standards at 3 to move ahead of older datasets that may have issues with the standards we define in this PR.

Open for discussion!


STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
-Current list:
-* Ensure all submissions have cmiles, most important are mapped hydrogen smiles
-* Ensure the WBO is requested for all submissions, this should be included in the scf properties list using the flag `wiberg_lowdin_indices`
+# Dataset Standards
 
-## Best practices
-* If any calculations are to be redone from another collection, re-use the old input (coordinates, atom ordering etc) as this will avoid running the calculation again and will just create new references in the database to the old results and should help keep the cost of the calculations down.  
+## Each molecule must have the following information:
+- Canonical isomeric explict hydrogen mapped SMILES
+- Provenence info of SMILES generation (NEW)
```
jthorton, line 13: I think this is definitely needed, but this will lead to a lot of repeated information as every molecule should have come through the same pipeline, so maybe it would be better to store this at a higher level like in the dataset metadata?

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+
+## Each dataset must have the following information:
+- Name
+- Version (NEW)
+- A short description
+- A long description
+- A link/URL/reference pointing to the provenence to reproduce dataset (e.g. the GH submission folder)
+- A changelog (NEW)
+- A github submitter username
+- The name of the person who selected/sourced the molecules (NEW)
```
jthorton, line 26: I agree that this is very useful information along with the changelog but is it not better to have this kind of detailed information in the README that the long description URL points to? I just want to make sure users don't have to repeat themself when making datasets making the process move involved as we are trying to make this as easy as possible.

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- A description of the meaning of the entry/molecule keys/names (NEW)
+- Each entry has canonical isomeric explict hydrogen mapped SMILES
+- Provenence info of CMILES generation (NEW)
```
jthorton, line 29: This would be the same information as the per molecule smiles provenance info.

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- A set of elements that the dataset contains
+- A set of charges that the dataset contains (NEW)
```
jthorton, line 31: Nice this will be very useful!

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- The mean and max molecular weight of molecules the dataset contains (NEW)
```
jthorton, line 32: Again this will be great!

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- Enumerated stereo flag (True/False) (NEW)
```
jthorton, line 33: I think these kinds of tags are really useful to get an idea at a glance of what went into making the dataset, and I think that this list of options has the potential to grow so maybe it would be better to have any, and all workflow stages used in QCSubmit printed here? The settings could also be printed but again this will be in the git repo here.

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+# Dataset Standards
 
-## Best practices
-* If any calculations are to be redone from another collection, re-use the old input (coordinates, atom ordering etc) as this will avoid running the calculation again and will just create new references in the database to the old results and should help keep the cost of the calculations down.  
+## Each molecule must have the following information:
+- Canonical isomeric explict hydrogen mapped SMILES
+- Provenence info of SMILES generation (NEW)
+- Coordinates
+- Total Charge
+- Coordinates must be in CMILES order (discuss)
```
jthorton, line 16: Yes this should always be true.

jthorton:
Thanks for putting this together @trevorgokey this is shaping up nicely! Some other comments on extra potential flags:
- maybe it would be good to include the intent of the dataset in the metadata such as benchmarking/fitting/exploratory/publication etc I know that Jessica and Pavan recently had trouble working out which datasets they could use in their work as they were not sure if it had been used in a fitting before?
- To further help with the point above we should restrict the names of datasets in the archive to not include benchmark/fitting and instead use tags which can be updated in the metadata or only allow our official benchmark/fitting datasets to use these names?
- A per molecule provenance field might be a good idea as well to indicate how a molecule was made for example a molecule made from enumerating the tautomers of another could have something like `provenance={"enumerated_tatutomers":"[H]c1c(c(nc(c1[H])C2=NN=N[N-]2)[H])[H]"}` here we would just track the operation that made the new molecule, this can also extend to how molecules were originally created capturing the toolkit method used to make them like `provenance={"created_with": "from_file"}`.

STANDARDS.md
trevorgokey: So, we should make the distinction now I guess. Should the CMILES in the extras field be mutable?

STANDARDS.md
trevorgokey: I am slightly tending towards having QCA as the primary source of data. I fear there would be a disconnect by storing informative pieces somewhere else. But, this could definitely be something that is voted on. 

In terms of technical aspects, I presume much of this (e.g. copying metadata for a new dataset) can be automated. For example, having e.g. QCSubmit have a "changes" field would just be some string users input, and the implementation can construct the full changelog. I think the tough issue is knowing which QCSubmit fields need to be filled out, but we can easily make a list/template notebook of this once these standards are settled.

STANDARDS.md
trevorgokey: Well, not necessarily given the 3D coordinates, but, this again asks whether CMILES info should be mutable QCA

STANDARDS.md
trevorgokey: Should we instead use heavy atoms as a count? I am trying to encapsulate the size of molecules that the dataset contains in a simple way

STANDARDS.md
trevorgokey: We prevent growing by explicitly stating what is in the standards :) Which ones do we want in this version? I am thinking stereo is stable and ready to use. I don't know whether tautomer is ready for official inclusion.

With stereo enumeration, I am trying to communicate patterns in the dataset. If I have a SMILES, does that mean it's enantiomer is also in the dataset somewhere? This flag indicates that. I don't know if indicating these were fragmented, for example, would convey any useful information in this regard (unless, of course, the dataset contained both the parent and the fragments, in which case a fragment flag would be useful).

Also, just putting all workflow components in this standard would mean forming a standards on filtering and that structure. I am thinking this could possibly come in a standards version later down the road.

STANDARDS.md
trevorgokey: This is not true.. What I mean here is that our SMILES is nearly always out of order with the 3D, hence why we have the mapped CMILES. We could possibly save headaches by reordering the 3D into the SMILES ordering prioer to submission, such that the mapped SMILES is always :1, :2, :3, :4, etc. We have things in the TK like `from_mapped_smiles`, but this doesn't quite work since the dihedrals index (for TDs) is still in the 3D index system. So, I presented this because, although I don't think it's necessary, it would make using the data much much easier for the general user.

trevorgokey:
Thanks for the input @jthorton! 

> maybe it would be good to include the intent of the dataset in the metadata such as benchmarking/fitting

I wasn't sure if it was worth including, but does this mean you'd like a more official name for benchmarking datasets as well? The current standards indicates a clear distinction for fitting datasets; perhaps we can do the same for benchmarking datasets. In my view, everything is exploratory except for the official fitting sets, so perhaps it is a good idea to distinguish official benchmark sets too.

> To further help with the point above we should restrict the names of datasets in the archive to not include benchmark/fitting and instead use tags which can be updated in the metadata or only allow our official benchmark/fitting datasets to use these names?

Yes, I think a clear naming scheme for fitting/benchmarking datasets is needed. I do like the idea of an explicit field somewhere that says "fitting" or "benchmarking" or "research". Perhaps a "purpose" field would be good in this case?

> A per molecule provenance field might be a good idea as well to indicate how a molecule was made for example a molecule made from enumerating the tautomers of another could have something like provenance={"enumerated_tatutomers":"[H]c1c(c(nc(c1[H])C2=NN=N[N-]2)[H])[H]"} here we would just track the operation that made the new molecule, this can also extend to how molecules were originally created capturing the toolkit method used to make them like provenance={"created_with": "from_file"}

So the last point aims at "where did we get the inputs needed to generate the inputs in the dataset?" I think this is complicated enough that we should just rely on the GH repo submission notebooks for that. I think the actual datasets should have a t=0 at the QCA input molecule, and not be required to keep track (in the QCA database) of how those molecules were generated. From my view of these standards, I just want to know "who/what generated the dataset, and what was the script needed to submit if someone else wants to reproduce the set (in their own server)?" For example, I would be happy with "QCSubmit prepared the dataset; here is the notebook to run it, and here are the files that it needs to run; here is the URL". I am not sure how to encapsulate that here without also defining standards for the submission process, which is out of scope.

I do like maintaining the enumeration relationship, such as what you described and/or having a list of all discovered enumerated CMILES in the entry for a given molecule (this would be nice to e.g. easily find the stereoisomers in a given dataset). However, I feel like this is more of a want than a need. Feel free to include it in future submissions, but I don't think it's a requirement to submit a dataset.

_That said_, if you can maybe provide a use case where this information is of high value, I can be persuaded :)

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
 
-Current list:
-* Ensure all submissions have cmiles, most important are mapped hydrogen smiles
-* Ensure the WBO is requested for all submissions, this should be included in the scf properties list using the flag `wiberg_lowdin_indices`
+# Dataset Standards
 
-## Best practices
-* If any calculations are to be redone from another collection, re-use the old input (coordinates, atom ordering etc) as this will avoid running the calculation again and will just create new references in the database to the old results and should help keep the cost of the calculations down.  
+## Each molecule must have the following information:
+- Canonical isomeric explict hydrogen mapped SMILES
```
j-wags, line 12: I'm thinking about an eventual future where we store CMILES _and_ InChI. This framing made me realize that we're expecting the CMILES in the current plan to do two things:

* Contain the graph molecule information that needs to be added to a QC representation to get something parameterizable.
* Be a lookup key for FINDING this particular molecule again in the future

I think CMILES is a really good tool for the first need. But I think it's bad at the second case (since chemically identical molecules could have different CMILES, so we have to record things like software provenance to make the lookup work). So, knowing that we may incorporate InChI in the future, we may not need to make our first attempt do this perfectly.

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
+	- OpenFF toolkit ingestion with strict stereo checking in RDKit
+	- Hydrogen bonding
+	- Torsion drives on rings or other high barrier issues
+	- CMILES change (no way to build new CMILES since connectivity cannot be trusted, also this means we can't trust per-molecule CMILES until this filtering check passes)
 
-* See ["Molecule submission checklist"](https://github.com/openforcefield/qcsubmit/issues/9)
+## Force Field Releases
 
-## Standard functions and modules for entry preparation
+Upon fitting for a new force field release, for the purpose of paper publication, public reference, etc, all molecules should be placed in a single dataset (per type). This gives a single reference for these data instead of many references. Filtering must be done prior, such that all molecules in the release datasets pass all post-submission filters.
```
j-wags, line 116: **Recommendation for immediate action:** I wouldn't be too specific about the filtering at this stage. I'd keep it at "do whatever you need, and record it in the fitting tarball, then when you're done, email me a list of the molecules that survived filtering and I'll make a new dataset from those"

**Recommendation for what we eventually want:** Like you said, it'd be great to codify the filtering rules in a QCSubmit pipeline, which produces a QCA dataset that's immediately ready for download + FB input (with no additional "offline" filtering needed).

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
+    - Cosmetic changes
+	- Updating the blacklist
+	- Updating the dataset status
 
-A tag `"inflight"` indicates that a dataset is not completed as far as OpenFF is concerned.
-This means that any errors remaining are being actively addressed.
+(It might be useful to make the version bump in the changelog, but not the actual dataset name; this would depend on the ability to changing a dataset metadata of an existing dataset. This would allow existing scripts to not need to update to use a new dataset name when e.g. we update the status to COMPLETE/DONE or NONCOMPLIANT/PAUSED->INCOMPLETE/WORKING).
 
-All datasets should also feature a `"openff"` tag.
+A best-effort is made to ensure that a dataset follows its underlying STANDARDS. One must assume that the newest version of a dataset best conforms to these STANDARDS, and the same promise may not hold for earlier versions. The changelog should address any changes made to improve compliance.
```
dotsdl, line 89: Great work on this Trevor!

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
-This means that any errors remaining are known to be acceptable or impossible to fix.
-It also means that no additional work is being done on the dataset to get it to completion.
+- A patch (e.g. `"v3.1.1"`) represents a cosmetic change, or a change that is based on dynamic information that does not change the underlying data:
+    - Cosmetic changes
+	- Updating the blacklist
+	- Updating the dataset status
 
-A tag `"inflight"` indicates that a dataset is not completed as far as OpenFF is concerned.
-This means that any errors remaining are being actively addressed.
+(It might be useful to make the version bump in the changelog, but not the actual dataset name; this would depend on the ability to changing a dataset metadata of an existing dataset. This would allow existing scripts to not need to update to use a new dataset name when e.g. we update the status to COMPLETE/DONE or NONCOMPLIANT/PAUSED->INCOMPLETE/WORKING).
```
dotsdl, line 87: Yeah, patch changes probably shouldn't be reflected in a dataset name, since many of these dynamic information bits make more sense as edit-in-place changes to mutable data. I think the best-effort paragraph below gives us enough operational room to handle the kinds of cases we need without being overly descriptive on the implementation path.

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
 
-All datasets should also feature a `"openff"` tag.
+A best-effort is made to ensure that a dataset follows its underlying STANDARDS. One must assume that the newest version of a dataset best conforms to these STANDARDS, and the same promise may not hold for earlier versions. The changelog should address any changes made to improve compliance.
 
-## Force Field Releases
+Each version increment should take the information from the previous `changelog` field, and add a new entry of the form { "version": description } that explains the modifications made to the dataset. Each dataset should therefore have the complete changelog.
```
dotsdl, line 91: I really like this. :heart:

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
 
-When a new force field is released, a dataset corresponding to all results used for the force field fitting should be created.
-This gives a single reference for these data instead of many references.
-The format of these dataset names is:
+# Fitting standards
 
-    `"OpenFF Force Field <friendly name> <ff version>"`
+Reference level of theory: `B3LYP/DZVP`
```
dotsdl, line 95: Perhaps change to `B3LYP-D3(BJ)/DZVP` to be more specific?

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
+Geometry optimizations: `geomeTRIC` using the TRIC coordinate system
+QM program: `Psi4`
```
dotsdl, line 97: Change to a markdown list for desired rendering, e.g.:
```suggestion
For fitting datasets, the following applies:
- Reference level of theory: `B3LYP/DZVP`
- Geometry optimizations: `geomeTRIC` using the TRIC coordinate system
- QM program: `Psi4`
```

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
 
-## Group
+For unconstrained geometries, all molecules must have:
+- Wiberg Bond Orders (parameter interpolation)
+- Hessian (frequency fitting) (is this too restrictive? discuss)
```
dotsdl, line 101: Are we currently storing Hessians only sometimes?

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
+- Unless explicitly specified in the submission descriptions, torsion drives must be on 4 connected atoms
+- Torsions driving a ring will give a warning, and torsions in a a ring of  3,4,5,6 atoms is considered an error
+- Warnings will be given if an atom does not have a complete valence set
 
-## Molecule validation
+Post-submission filtering: (we need testers/validators for these)
+	- OpenFF toolkit ingestion with strict stereo checking in RDKit
+	- Hydrogen bonding
+	- Torsion drives on rings or other high barrier issues
+	- CMILES change (no way to build new CMILES since connectivity cannot be trusted, also this means we can't trust per-molecule CMILES until this filtering check passes)
```
dotsdl, line 112: Remove indentation for markdown rendering.

dotsdl: Implementation thought: this validation could be built into QCSubmit as part of our dataset access layer, generally used before using data for a fitting cycle to weed out "bad" candidates for fitting?

STANDARDS.md
dotsdl, line 116: Ah, this is very nice! This closes the loop for fitting datasets. I can see how this fits together with the post-submission filters you recommended above.

STANDARDS.md
```diff
@@ -19,54 +69,60 @@ Each dataset shall be versioned.
...
+    `"OpenFF SMIRNOFF <friendly name> <ff version>"`
 
-### Required fields
+for example, all datasets (optimizations, torsion drives, and Hessians) with the name `"OpenFF SMIRNOFF Sage 2.0.0"` would refer to all data used to train `openff-2.0.0.offxml` in the `openforcefields` package.
 
-* See ["Fields that should be required for OpenFF submissions"](https://github.com/openforcefield/qcsubmit/issues/3)
+Besides the regular information from the other datasets, these fitting datasets must have:
 
+- `DOI`
+- Dataset standards version (this document)
```
dotsdl, line 127: Awesome. This is some excellent provenance thinking! Great work Trevor!

STANDARDS.md
dotsdl: Does InChi "solve" this issue (the same InChi always produced for chemically-identical molecules), or is it just a different system from mapped CMILES with its own (but different) edge cases?

STANDARDS.md
dotsdl: Could be a precendence model, where top-level dataset metadata applies to all molecules *unless* that molecule has its own, perhaps conflicting metadata; finest scope (molecule-specific) wins where present.

dotsdl (commented):
Awesome work @trevorgokey! You've built an incredibly functional standards policy document here, and I think we're already at something that could be adopted. This will guide our advancements in submission machinery, implementing the policies here, which in turn will result in higher-quality datasets that are useful to others for a long time.

I'd like to push for adoption within this next week if possible. Can you respond to or adopt suggestions from feedback comments here, and then present at the submission meeting on Friday 12/4 or Friday 12/11? If you are ready, we can adopt these STANDARDS and merge after some final discussion and laying of hands.

hyejang:
@trevorgokey does this set of standards contain (or will contain) a guideline on how to add new molecules/entries into the existing dataset?

STANDARDS.md
trevorgokey: I am ok with this, but I think it could complicate things. For example, if I am a user and I see that provenance exists for one molecule but not another, is this a bug or expected behavior? What logic would I need to handle this? I would have to discover this in documentation somewhere that missing data means more top-level structures (e.g. entries) take precedence. My opinion here is that the data should be completely transparent (e.g. if it exists for one it should exist for all), or the actual QCA client provides a consistent interface and does the magic to find the right field if we have a sparser, more efficient representation. I am not a fan of having a sort of custom API layer outside of the client that people *must* use because we have enough tricks in the way we structure our data that it becomes unusable to non-experts.

One other issue is that a molecule can have multiple entries associated with it, e.g. it can belong in multiple datasets. In this case, if provenance is undefined for the molecule CMILES, which entry should I assume the provenance is from? We could apply (likely undocumented) logic here, but this could be avoided if we just applied provenance everywhere the function was used.

STANDARDS.md
trevorgokey: I agree with this, but there are complications that have prevented this. For example, QCA does not have the notion of data precedence, which is why we cannot for example take CMILES info from an entry and instead need to store it in the molecule. If there was a precedence model in place, then the molecule wouldn't need CMILES info in the first place.

STANDARDS.md
trevorgokey: Molecule weight involves isotopes, and as far as I know these shouldn't affect most of our calculations. I agree that we could do both for the sake of doing both. On the QM side, I guess number of electrons is more useful, but on the MM/informatics side, I am thinking weight could be a better descriptor here.

STANDARDS.md
trevorgokey: @j-wags could you possibly comment on perspectives and issues that may warrant not including stereo and tautomer enumeration flags in the datasets?

STANDARDS.md
trevorgokey: Yes, I think this would be the implementation of my note, but the question is... should we do/enforce this? We already have an unmapped version (`canonical_isomeric_explicit_hydrogen_smiles`), which would also mean that the ordering of these atoms is the ordering of the 3d coordinates.

jthorton: Oh I mean the coordinates should always be in the order of the mapped smiles which is always true. This problem is avoided if we always make a molecule from the mapped smiles, this means you can just add a conformer to the molecule without having to do any reordering, the dihedral array in record is then correct and matches the molecule. We could, of course, have QCSubmit enforce canonical ordering of the molecule however to avoid confusion.

STANDARDS.md
trevorgokey: In the name of getting this document through, I will hold off on this since it would require extra implementation work and has less value per person-hour than other issues. I am a big of this working, so I am hopeful that it gets adopted in later versions.

STANDARDS.md
jthorton: Yes it should solve this as inchi is a separate standard that all toolkits agree on and I think it would be great to include in the molecule as well.

STANDARDS.md
trevorgokey: Alright, for this release we can keep things closer to what is already being done. I do hope that these pipelines become better defined in the near future (i.e. the subsequent version of this document).

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- The changelog must be copied from the current version, and a new entry added in the new version
+- The dataset version information is updated
+- The README is updated
+- A notebook or record of the lines of python used to manipulate the dataset responsible for the revision
+	- If the revision should have the new version in the filename e.g. `update-v3.1.1.py` (this does not work for notebooks; discuss)
+- Update the `index` of datasets on the GH repo
+
+## QCArchive-specific dataset standards
+
+- Compute tag is `openff`
```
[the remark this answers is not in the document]
trevorgokey: I agree that having multiple tags is a valuable asset, but I am not a fan of leaving this in a gray zone due to devops/technical issues. Can you explicitly state which tags are possible and in use, so it is more transparent?

STANDARDS.md
```diff
@@ -1,16 +1,66 @@
...
+- Each entry has canonical isomeric explict hydrogen mapped SMILES
+- Provenence info of CMILES generation (NEW)
+- A set of elements that the dataset contains
+- A set of charges that the dataset contains (NEW)
+- The mean and max molecular weight of molecules the dataset contains (NEW)
+- Enumerated stereo flag (True/False) (NEW)
+- Enumerated tautomers flag (True/False) (NEW)
+- Enumeration provenence info (NEW)
+- Computation blacklist (known failures) (NEW)
+* Dataset status (NEW):
```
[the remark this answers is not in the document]
trevorgokey: Sure! I especially like the compliance status format.

STANDARDS.md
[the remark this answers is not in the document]
jthorton: I agree that the tags should be defined but maybe they don't have to be in the standards, I think we probably want some docs for this repo that would have a list of them and is updated regularly?

STANDARDS.md
trevorgokey: Yes, that is somewhat the idea with filtering standards. However, this will get passed on in the current version for now.

STANDARDS.md
trevorgokey: However, I do think this is valuable enough to leave into the standards; it just may mean that we take longer to implement this document.

STANDARDS.md
trevorgokey: I think having the compliance status with an explicit `V3` status will also help with this nicely.

trevorgokey:
@hyejang This document definitely supports this route, but doesn't quite touch on _how_ to do it as that is an implementation detail. All we require is following the steps outlined in revision creation procedure. We hope this document can be implemented ASAP, with most things automated so users don't need to worry about most of it.

dotsdl:
Thanks @trevorgokey for putting this together! In today's QCFractal Submission meeting, we resolved to adopt these STANDARDSv3! 

@jthorton has committed to translating the components expressed in these STANDARDS that require implementation yet as issues on QCSubmit and qca-dataset-submission so we can coordinate development effort!

dotsdl (approved):
This document has been adopted!

PR #139, opened 2020-10-01
Updated prod envs to use qcfractal 0.14.0, latest openff packages

dotsdl:
This PR updates all prod environments to use `qcfractal 0.14.0` explicitly, as well as pinning to the latest versions of `openforcefield`, `openforcefields` where used.

A QCEngine release may be imminent, so we will wait for word on this before merge and deployment. That will give us improvements for both the Psi4 and ANI harnesses that we really want for production use.

### DIFF
--- a/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
+++ b/devtools/prod-envs/qcarchive-worker-openff-ani.yaml
@@ -6,7 +6,7 @@ channels:
 dependencies:
   - python =3.7
   - pip
-  - qcfractal
+  - qcfractal =0.14.0
   - qcengine =0.16.0
 
   # ML calculations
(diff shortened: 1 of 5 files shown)


dotsdl:
@trevorgokey please have a look to see if I've missed anything, or if there's something we shouldn't do here. If you get the chance to test these files out locally, that would also be very useful. Thanks!

trevorgokey:
Looks good @dotsdl! I submitted a torsiondrive to a local server with MM and QM specs and everything is running smoothly.

trevorgokey (approved):

dotsdl:
Awesome, thank you @trevorgokey! This reduces my blood pressure substantially. :sweat_smile:

dotsdl:
Added `qcarchive-worker-openff-xtb`, with a version pin to the latest `xtb-python` (20.1). The `xtb-python` conda package doesn't pin to a particular version of `xtb` itself, but we probably shouldn't pin that explicitly unless we encounter issues. 

Do these environments work for your usage of `xtb` @jthorton?

dotsdl:
QCEngine 0.17.0 is now up on conda-forge. These should be ready to merge whenever convenient.

PR #140, opened 2020-10-02
Added supplemental compute spec to OpenFF Substituted Phenyl Set 1

dotsdl:
This PR expands the existing "OpenFF Substituted Phenyl Set 1" TorsionDrive dataset to include additional compute specs. We include the following here:
- default (already computed)
- openff-1.0.0
- openff-1.2.0
- openff-1.2.1
- gaff-2.11

----
- [ ] Ready to submit!

### DIFF
--- a/submissions/2019-07-25-phenyl-set/compute.json
+++ b/submissions/2019-07-25-phenyl-set/compute.json
@@ -54,10 +54,10 @@
     "submitter": "dotsdl",
     "creation_date": "2020-10-02",
     "collection_type": "TorsiondriveDataset",
-    "dataset_name": "TorsionDriveDataset",
-    "short_description": "OpenForcefield TorsionDrives.",
+    "dataset_name": "OpenFF Substituted Phenyl Set 1",
+    "short_description": "TorsionDrives for WBOs, torsion relationship",
     "long_description_url": null,
-    "long_description": "A TorsionDrive dataset using geometric.",
+    "long_description": "A TorsionDrive dataset using geometric; investigating WBO and torsion barrier relationship",
     "elements": []
   },
   "provenance": {},
--- a/submissions/2019-07-25-phenyl-set/compute.json
+++ b/submissions/2019-07-25-phenyl-set/compute.json
@@ -1,0 +1,95 @@
+{
+  "qc_specifications": {
+    "default": {
+      "method": "B3LYP-D3BJ",
+      "basis": "DZVP",
+      "program": "psi4",
+      "spec_name": "default",
+      "spec_description": "Standard OpenFF optimization quantum chemistry specification.",
+      "store_wavefunction": "none"
+    },
+    "openff-1.0.0": {
+      "method": "openff-1.0.0",
+      "basis": "smirnoff",
+      "program": "openmm",
+      "spec_name": "openff-1.0.0",
+      "spec_description": "default openff-1.0.0 spec",
+      "store_wavefunction": "none"
+    },
+    "openff-1.2.0": {
+      "method": "openff-1.2.0",
+      "basis": "smirnoff",
+      "program": "openmm",
+      "spec_name": "openff-1.2.0",
+      "spec_description": "default openff-1.2.0 spec",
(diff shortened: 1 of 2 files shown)


dotsdl:
Ah, we get an X on "Validated" if the spec is already present. I'll remove the `default` spec since it's there. Really thankful we have these. :grin:

dotsdl:
@trevorgokey do you mind giving me a quick review of this if validations pass? We'll then be clear for submission.

submissions/2019-07-25-phenyl-set/compute.json
```diff
@@ -0,0 +1,87 @@
...
+    "wiberg_lowdin_indices",
+    "mayer_indices"
+  ],
+  "priority": "normal",
+  "description": "A TorsionDrive dataset using geometric.",
+  "dataset_tags": [
+    "openff"
+  ],
+  "compute_tag": "openff",
+  "metadata": {
```
trevorgokey, line 53: Would like to see the metadata look a little more complete/representative. Was this generated from scratch, or pulled from the existing dataset?

submissions/2019-07-25-phenyl-set/compute.json
```diff
@@ -0,0 +1,87 @@
...
+    "submitter": "dotsdl",
+    "creation_date": "2020-10-02",
+    "collection_type": "TorsiondriveDataset",
+    "dataset_name": "TorsionDriveDataset",
+    "short_description": "OpenForcefield TorsionDrives.",
+    "long_description_url": null,
+    "long_description": "A TorsionDrive dataset using geometric.",
+    "elements": []
+  },
+  "provenance": {},
```
trevorgokey, line 63: Likely QCSubmit provenance should be put here

submissions/2019-07-25-phenyl-set/compute.json
dotsdl: I pulled this JSON from the `compute.json` of the Benchmark Ligands set. I modified it where needed. What would you like here?

submissions/2019-07-25-phenyl-set/compute.json
dotsdl: I think QCSubmit might inject some? Not sure exactly. @jthorton?

dotsdl: Ah nevermind, this normally gets populated when QCSubmit is used to generate, say, `dataset.json`. "provenance" indicates what version of `qcsubmit`, `openforcefield`, and `openeye` were used to generate it. Since this `compute.json` isn't generated from use of QCSubmit here, I think it's fine to leave it blank.

submissions/2019-07-25-phenyl-set/compute.json
dotsdl: Added some more specific metadata. Thanks for pointing this out @trevorgokey!

jthorton:
@dotsdl as the dataset is old it has no cmiles in the molecules and is taking the RDKit path to parameterize the molecules in QCEngine so the errors will not clear. If the other molecules results are needed then the dataset will have to remade using QCSubmit so the MM scans can be run.

dotsdl:
> @dotsdl as the dataset is old it has no cmiles in the molecules and is taking the RDKit path to parameterize the molecules in QCEngine so the errors will not clear. If the other molecules results are needed then the dataset will have to remade using QCSubmit so the MM scans can be run.

@jmaat are these results usable to you despite failures noted by @jthorton above? Even if so, I can prepare a new dataset matching the original, but with metadata needed high-fidelity QM->MM translation instead of guessing with RDKit. Please confirm if this is something you'd like.

jmaat:
@dotsdl Yes, it's fine to proceed with the MM energies for this dataset

jmaat:
@dotsdl Just want to make sure these torsion drives are using 1.2 and 1.3 OpenFF

dotsdl:
@jmaat the compute spec will get submitted from #161. I've added OpenFF 1.3.0 to the list. :grin:

PR #141, opened 2020-10-06
OpenFF Protein Peptide Fragments constrained v1.0

jthorton:
- [x] Ready to submit!

This is the protein fragment optimization dataset remade with corrected indices. Please do not merge yet I want to double-check that the constraints are definitely correct before running.


dotsdl:
Ah, just saw your note @jthorton to not merge yet. I'll hold off on any further adjustments. Let me know if you'd like me to revert anything.

jthorton:
@dotsdl Thanks for making those changes I was worried about removing the old dataset completely but I guess it is in the git history. I just want to check once more tomorrow that the constraints are correct then I'll submit if you are happy with it.

dotsdl:
Excellent, sounds good @jthorton! I looked for the `int(x) - 1` in the constraint gatherer, and if your sanity checks show these to be the right atoms for a set of molecules, looks good to me!

I just did a manual sanity check comparing the constraints for GLY_CYS_ALA against the corresponding `constraints` in the `dataset.json` submission file. I see that each dihedral constraint in `dataset.json` for this case have a 1:1 match with the dihedral constraint lines in `Input_files/CYS/GLY_CYS_ALA/Conf115.frz`, decremented by 1. Can confirm that those constraint indices in the FRZ file correspond to the correct heavy atom identities in the corresponding MOL2/PDB file.

jthorton:
Thanks for checking over them @dotsdl I also added a check in the notebook which uses the torsion validator from qcsubmit to make sure the constraints are on valid torsions. It looks like constraint validators would pick this up in the future I have made an issue [here](https://github.com/openforcefield/qcsubmit/issues/45) to collect some thoughts. Overall I think this is ready to go I just need to update QCsubmit with compression/decompression so the check works.

dotsdl:
Ah sorry @jthorton, we haven't included xz decompression in the shim. I could add it if you like?

jthorton:
No problem I can recompress tomorrow using another format.

dotsdl:
@trevorgokey I know your compute is starving; if you could give this a review, I think @jthorton would be fine with launching it?

submissions/2020-08-12-OpenFF-Protein-Fragments-version2/README.md
```diff
@@ -1,32 +1,54 @@
...
-### General Information
+Details from David:
+
+> The way it's set up, we are scanning phi and psi of the central residue with a random selection of ALA, GLY, SER, or VAL to the N- or C-terminus, and the customary ACE and NME blocking groups outside of that.
+
+## General Information
+
  - Date: 12/8/2020
  - Class: OpenFF constrained optimization 
  - Purpose: Optimizations
```
trevorgokey, line 14: Please make this more descriptive.

submissions/2020-08-12-OpenFF-Protein-Fragments-version2/README.md
```diff
@@ -1,32 +1,54 @@
...
+ - Submitter: Josh Horton
+ 
+## Changelog
+
+Here any information regarding dataset changes are recorded.
+
+### v2.1
+
+The dataset constraints were found to be incorrect due to a mistake in the indexing.
+The indices were assumed to be 0 based but were indexed from 1; this is corrected in version 2.1.
```
trevorgokey, line 27: Which indices are these referring to? The provided inputs?

submissions/2020-08-12-OpenFF-Protein-Fragments-version2/README.md
```diff
@@ -1,32 +1,54 @@
...
+ - Date: 2020.10.06
+ - Class: OpenFF constrained optimization 
+ - Purpose: Optimizations 
+ - Collection: OptimizationDataset
+ - Name: OpenFF Constrained Protein Fragments v2.1
+ - Number of unique molecules: 185
  - Submitter: Josh Horton
  
 Note, each folder contains molecules saved via mol2 in each confirmation however the bond order is incorrect, we let openeye interpret it by re-saving to PDB first before creating the dataset.
```
trevorgokey, line 36: Is there a record of how these mol2 files were generated, and why proper inputs weren't provided from the start? I am not excited about having a piece of software reinterpret the entire input, since this can be a very real source of issues

trevorgokey (commented):
Some comments on the notebook, since GH does not provide an easy way to comment on them. Please clean it up a bit; including large print outs of data. Also, it is preferred that the submission notebook be free of errors/raised exceptions/etc. It is nice to show that the QCSubmit throws an exception on bad input, but that is a unit test for QCSubmit rather than an important piece for the submission of this dataset. The notebook should simply walk me through the steps needed to submit the molecules.

Lastly, if I am being super pedantic, I would ask for a PDF of the molecules with all constraints highlighted. I currently have no current way to easily confirm that the constraint indices are valid. I assume that they were correct based on the primary source of input, so the fact that there are errors means this PR is modifying the molecules in ways that are producing issues. It would be a clear solution to prove/show that the molecules after modification by this PR are consistent with the input, which is not shown/verified anywhere other than checking that they are connected, which does not completely show that the original constraints are respected.

trevorgokey:
Also, please modify the long description in the submitted metadata to not be relative to a previous version. The descriptions should be standalone and completely describe everything in the current set.

submissions/2020-08-12-OpenFF-Protein-Fragments-version2/README.md
jthorton: Yes these are directly from mdgx I believe.

submissions/2020-08-12-OpenFF-Protein-Fragments-version2/README.md
jthorton: Yes that's correct the raw inputs were assumed to be zero-based which caused the confusion.

trevorgokey (approved):
Looks beautiful! Thanks for all of the changes @jthorton. 

My only note is that since the connected torsions were not explicitly shown in the notebook, that they are automatically done in one of the QCSubmit calls. The fact that no exception is raised, that the test passed.

As a general note for posterity: a PDF of the constraints were generated offline, and they appear valid. So, we believe the constraints should be correct at this point.

PR #143, opened 2020-10-07
Added Optimization regeneration for stalled INCOMPLETEs

dotsdl:
Adds regeneration of hanging INCOMPLETE tasks to automated error cycling.

### DIFF
--- a/management/management.py
+++ b/management/management.py
@@ -176,6 +176,13 @@ def restart_optimizations(optimizations, client):
             client.modify_tasks(operation='restart', base_result=opt.id)
 
 
+def regenerate_optimizations(optimizations, client):
+    for opt in optimizations:
+        if opt.status == 'INCOMPLETE' and (opt.trajectory is not None):
+            print(opt)
+            client.modify_tasks(operation='regenerate', base_result=opt.id)
+
+
 def restart_torsiondrives(torsiondrives, client):
     for tdr in torsiondrives:
         if tdr.status == 'ERROR':
--- a/management/management.py
+++ b/management/management.py
@@ -172,19 +172,19 @@ def count_unique_optimization_error_messages(
 def restart_optimizations(optimizations, client):
     for opt in optimizations:
         if opt.status == 'ERROR':
-            print(opt)
+            print(f"Restarted ERRORed optimization `{opt.id}`")
             client.modify_tasks(operation='restart', base_result=opt.id)
 
 
 def regenerate_optimizations(optimizations, client):
     for opt in optimizations:
-        if opt.status == 'INCOMPLETE' and (opt.trajectory is not None):
-            print(opt)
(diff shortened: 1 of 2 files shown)


dotsdl:
@trevorgokey can I get your thoughts on the wisdom of this approach?

management/management.py
```diff
@@ -176,6 +176,13 @@ def restart_optimizations(optimizations, client):
             client.modify_tasks(operation='restart', base_result=opt.id)
 
 
+def regenerate_optimizations(optimizations, client):
+    for opt in optimizations:
+        if opt.status == 'INCOMPLETE' and (opt.trajectory is not None):
```
trevorgokey, line 181: I am not quite sure what would constitute a satisfying requirement for regeneration, and my gut feeling tells me something will slip through here. Also, it seems like this will try to regenerate tasks that are currently running by a manager. We can possibly be a bit more careful by checking if `final_molecule` is populated, as that is a better indicator that an optimization really finished.

management/management.py
```diff
@@ -176,6 +176,13 @@ def restart_optimizations(optimizations, client):
...
+            print(opt)
+            client.modify_tasks(operation='regenerate', base_result=opt.id)
```
trevorgokey, line 183: Given that we don't know how well this will work, it might be helpful to print some context with this optimization, like "this optimization ID ### was regenerated"

management/management.py
```diff
@@ -176,6 +176,13 @@ def restart_optimizations(optimizations, client):
             client.modify_tasks(operation='restart', base_result=opt.id)
```
trevorgokey, line 176: Are we sure that regenerated tasks are automatically eligible for submission? Should we try to restart after regen?

trevorgokey (changes requested):

management/management.py
dotsdl: Ah, agreed! Will switch out `opt.trajectory` check with a check on `final_molecule`.

management/management.py
dotsdl: Will add additional context, good call!

management/management.py
dotsdl: I did an ad-hoc test with the old Fragmenter dataset we are trying to get clear of INCOMPLETEs. I'm not sure it had any effect. Posed a more targeted question in slack.

dotsdl: Looks like it cleared from Friday's discussion!

management/management.py
dotsdl: Done! Thanks Trevor!

management/management.py
dotsdl: Done!

dotsdl:
@trevorgokey I've made your requested changes. Is this good to go?

trevorgokey (approved):
Looks good!

dotsdl:
Thanks @trevorgokey! Merged!

PR #145, opened 2020-10-09
[WIP] OpenFF Benchmark Ligands - Unfragmented v1.0

dotsdl:
This submission will perform TorsionDrives on the the JACS molecule dataset, without fragmenting the molecules, using only MM and ML potentials, namely:
- `openff 1.0.0`
- `openff 1.1.0`
- `openff 1.2.0`
- `openff 1.2.1`
- `ani2x`

In order to eliminate known errors for ANI2x, molecules with elements outside its supported set of H, C, N, O, S, F, Cl have been excluded.

This set was requested by @dominicrufa. 

## Launch Checks

- [ ] Ready to submit!


dotsdl:
I see some cases like the below in `molecules.pdf` where it appears that only 3 atoms are highlighted. @jthorton, @trevorgokey, any ideas what might be happening here?

Screenshot from 2020-10-12 12-46-34

dotsdl:
@jthorton I believe you had mentioned a parameter we might try altering for the optimizations since for the [fragmented version of this dataset](https://github.com/openforcefield/qca-dataset-submission/pull/136#issuecomment-703758881) we saw many failures due to lack of convergence on the part of ANI2x. I *think* the suggestion was to loosen the convergence tolerance for this case?

dotsdl:
@trevorgokey or @jthorton, would one of you kindly give this a review? I also had some questions above we'll want to address before we merge. Thank you!

trevorgokey:
Where was the logic defined on which torsions to drive? It is unspecified anywhere in this submission other than the PDF as I can't find any record in the notebook. I see a lot of methyls being driven, which need to be removed. I suspect that the torsion with 3 atoms in the visual is using a hydrogen in the definition, which should also be removed.

trevorgokey (changes requested):
Where was the logic defined on which torsions to drive? It is unspecified anywhere in this submission other than the PDF as I can't find any record in the notebook. I see a lot of methyls being driven, which need to be removed. I suspect that the torsion with 3 atoms in the visual is using a hydrogen in the definition, which should also be removed.

Also, given the situation of ANI, I would be delighted if someone could pull a TD that had optimization issues, and check what actually prevented it from converging. Was it a disaster, or was it off by 1e-4? I think this should be examined before we play guess and check with the settings. Just looking at the TD trajectory should be valuable; the geometric output should also be a good indicator or whether it was just slightly bouncy or ricocheting

jthorton:
Just to clear up what we are seeing here when no specific method is used to highlight torsions for driving the torsiondrive factory in QCSubmit is selecting every rotatable bond which is probably not needed as we see many methly, alcohol and amino bonds being rotated here, I will update QCSubmit to avoid this in future this is also the cause of the 3 atom highlights in some of the molecules in the PDF as a hydrogen is the 4th atom but they are not explicitly drawn.

trevorgokey:
@jthorton I think it is fine to use all rotatable bonds if nothing is specified. We could add options to turn of H driving and ring driving; those should be valuable. In this submission it is hard to tell what it did by looking at the output generated. Perhaps the settings/options on how TDs were generated should be more explicitly stated somewhere (even in the README for now), or even included as a workflow component which can hook into the richer provenance and option settings the components have to offer.

dotsdl:
Thank you both for this feedback. I will do the following:

1. Assist @dominicrufa to analyze the behavior of a set of convergence-failed TorsionDrive Optimizations. We will start with visualizing the trajectory, and from those observations try out different optimization parameters to see if we can arrive at a successful run. This will inform our approach here; I agree that this information will increase our chances of a usable submission.

2. I will check the central bond in each molecule's specified dihedral for a match with e.g. `[*]~[*:1]-[X2H1,X3H2,X4H3:2]-[#1]` (suggested by @jthorton). I will pop these out of the dataset as we find them before writing to `dataset.json`.

dotsdl:
Closing, as we have submitted #154 and are preparing #155 to fulfill the purpose of this dataset.

PR #149, opened 2020-10-19
Pinning ambertools to 20.4 for prod envs with MM

dotsdl:
Currently we get the 20.9 `ambertools` conda package for the `qcarchive-worker-openff-openmm` environment, but 17.0 for the `qcarchive-worker-openff` env with "everything".

The pinning added in this PR yields the 17.0 conda package consistently. Confusingly, this appears to still give us the version 20.0 of tools such as `antechamber`. We may desire consistency over anything else here; discussion welcome!


trevorgokey (approved):
Yeah, let's pin it for now to prevent unexpected surprises.

PR #153, opened 2020-10-30
OpenFF BCC Refit Study COH v1.0

SimonBoothroyd:
- [x] Ready to submit!


trevorgokey:
As an update; we might be able to unblock this, including not relying on software updates, as long as we are able to handle the data that the current wavefunction protocols provide. We will need to confirm that, roughly, `Da = C[:,nalpha]*C[:,nalpha].T`. Currently we are waiting/trying/seeking to store `Da`, but current working versions store `C`, which should be enough as long as everything looks correct.

jthorton:
As this dataset will also need a PCM spec latter we should not submit this until QCSubmit is updated to force c1 symmetry!

dotsdl:
@jthorton can you re-roll this dataset with the changes it needs? Should be ready for submission!

dotsdl:
@trevorgokey I think this one is ready for a final review!

SimonBoothroyd:
> As this dataset will also need a PCM spec latter we should not submit this until QCSubmit is updated to force c1 symmetry!

@jthorton is this in place now?

jthorton (approved):

trevorgokey (approved):
Everything looks good! Thanks @jthorton for giving this another check

dotsdl:
@SimonBoothroyd I cannot get the last `resp2-water` result to converge successfully, even with 2000 SCF iterations. Is this vital?

PR #154, opened 2020-10-30
[2020-07-27-OpenFF-Benchmark-Ligands] add a new ANI2X spec with more iterations

trevorgokey:
Adds a new `compute_ani2x_maxiter.json` file with a new ANI2X spec with more iterations. This is in effort to examine if we can get more successful calculations, since the current spec with 300 iterations produced many failures.

Everything in this update is the same otherwise, including molecules, etc.

- [x] Added new compute specification
- [x] Updated README

### DIFF
--- a/submissions/2020-07-27-OpenFF-Benchmark-Ligands/README.md
+++ b/submissions/2020-07-27-OpenFF-Benchmark-Ligands/README.md
@@ -52,6 +52,11 @@ This is a torsiondrive dataset created from the [OpenFF FEP benchmark dataset](h
         - method: ani2x
         - basis: None
         - program: torchani
+    - ani2x-v2:
+        - name: ani2x
+        - method: ani2x
+        - basis: None
+        - program: torchani
     - gfn2xtb:
         - name: gfn2xtb
         - method: gfn2xtb
--- a/submissions/2020-07-27-OpenFF-Benchmark-Ligands/README.md
+++ b/submissions/2020-07-27-OpenFF-Benchmark-Ligands/README.md
@@ -53,7 +53,7 @@ This is a torsiondrive dataset created from the [OpenFF FEP benchmark dataset](h
         - basis: None
         - program: torchani
     - ani2x-v2:
-        - name: ani2x
+        - name: ani2x-v2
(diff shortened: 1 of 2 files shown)


dotsdl (approved):
Looks good @trevorgokey! Thanks for putting this together. It will be interesting to see how well this expansion of the iterations allowed for convergence with ANI2x impact our success rates.

jthorton:
Just adding a note here I expect the maximum number of torsiondrives which can complete due to full element coverage to be 439/481.

dotsdl:
@jchodera we don't currently have any managers executing ANI2x, to my knowledge, since we sprinted on PEPCONF. This compute spec therefore hasn't made additional progress. Can you spin up some ANI2x workers, and I will do so as well?

jchodera:
> @jchodera we don't currently have any managers executing ANI2x, to my knowledge, since we sprinted on PEPCONF. This compute spec therefore hasn't made additional progress. Can you spin up some ANI2x workers, and I will do so as well?

Apologies @dotsdl --- I had missed this! I'll see if I can spin up some "kitchen sink" workers.

jchodera:
I'm still a bit unclear as to what is going on here. The original idea was that we could complete torsion drives with ANI2x in a few seconds. This has been running for months, and we're no closer to completing the remaining 50% of the dataset. Something is horribly, pathologically wrong here.

Is this the same "texturing" convergence issue that we ran into earlier? If so, we should probably terminate this dataset until Adrian and Oles come out with an updated version of ANI2x that is twice continuously differentiable.

jthorton:
> Is this the same "texturing" convergence issue that we ran into earlier? If so, we should probably terminate this dataset until Adrian and Oles come out with an updated version of ANI2x that is twice continuously differentiable.

Correct this was our attempt to fix the convergence issues we were seeing before we identified this problem, here we just upped the max optimization steps. Ill stop this now until we have a new model to test.

jchodera:
It sounds like Adrian Roitberg is about ready to release the new continuously-differentiable ANI2x variant. Hopefully this will be available through `torchani` shortly!

PR #155, opened 2020-11-03
Benchmark ligand ani1x compute

jthorton:
- [ ] Ready to submit!

This adds compute to the benchmark ligand fragments torsiondrives for ani1x and ani1ccx with a high maxiter. Due to the restricted coverage of ani1x I expect the maximum number of torsiondrives which we can complete to be 246/481.


dotsdl:
Letting this sit for now until we make more progress on #154.

PR #160, opened 2020-11-12
Added MM compute specs to CHO sandbox dataset

dotsdl:
- [x] Ready to submit!

### DIFF
--- a/submissions/2020-09-18-OpenFF-Sandbox-CHO-PhAlkEthOH/compute.json
+++ b/submissions/2020-09-18-OpenFF-Sandbox-CHO-PhAlkEthOH/compute.json
@@ -1,0 +1,105 @@
+{
+  "qc_specifications": {
+    "openff-1.0.0": {
+      "method": "openff-1.0.0",
+      "basis": "smirnoff",
+      "program": "openmm",
+      "spec_name": "openff-1.0.0",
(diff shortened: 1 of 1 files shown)


dotsdl:
This compute is ready for review! I used the same set of MM compute we are planning to use for benchmarking here.

jthorton (approved):

dotsdl:
@jthorton I think we'll need to be able to pass through `overwrite=True` to the submit method for a dataset. We're getting failure here because on merge the first submission took longer than the 6 hour max allowed for a GH Action job. In order to try again, we'll need to be able to `ds.submit(overwrite=True)` in lifecycle, I believe.

jthorton:
It looks like all of the specs were added but no tasks were generated maybe the compute call timed out? I am not sure if we want the ability to overwrite specs as this would lead to the db inconsistency that ben mentioned. I think in this case we just want to check if the spec we are adding is already there and if it is then run the compute call else throw an error to tell the user that this spec is already present and does not match so it needs to be added with a new name?

jthorton:
@dotsdl This QCSubmit issue is now fixed, however, we still see a time out when generating the tasks ~36k tasks in this submission I think the only way to get this through is to split each MM compute spec into separate PRs/actions.

PR #158, opened 2020-11-09
OpenFF Substituted Phenyl Set 1 v2.0

dotsdl:
This is a resubmission of the `OpenFF Substituted Phenyl Set 1 v1.0`, in response to metadata not being [stored in the places we need](https://github.com/openforcefield/qca-dataset-submission/pull/140#issuecomment-705675738) to use the dataset for MM reliably.

I have:
1. Used the original conformers, with no interpretation or conformer generation with the `openforcefield` toolkit. If you are reviewing, can you please confirm the approach taken matches this intention?
2. Removed molecules that we could not get to SCF-converge in the original set. These all include iodine, though other iodine-including molecules did not fail to SCF-converge. I tried to get these molecules to converge with up to `maxiter=3200` on local infrastructure to no avail. These molecules are:
    - `CC[CH2:4][NH:3][c:2]1ccnc([cH:1]1)I`
    - `C[C:4](=O)[NH:3][c:2]1[cH:1]cnc(c1)I`
    - `[CH3:4][O:3][c:2]1[cH:1]cnc(c1)I`

----

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added README.md describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset is named `dataset.json`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Ready to submit!

PR #161, opened 2020-11-12
MM compute for OpenFF Substituted Phenyl Set 1 v2.0

dotsdl:
This branch is based on the PR for #158, and should only be merged after that one is.

----

- [ ] Ready to submit!


submissions/2020-10-06-OpenFF-Phenyl-Set/compute.json
```diff
@@ -0,0 +1,87 @@
+{
+  "qc_specifications": {
+    "openff-1.0.0": {
+      "method": "openff-1.0.0",
```
jthorton, line 4: I think for openff spec we now want to use `openff_unconstrained-1.0.0` even though the constraints are automatically removed from any forcefield that goes through the new openmm harness this just makes it clear to the user of the data, or maybe we could put in the spec description that it is unconstrained? @trevorgokey?

jthorton (approved):
This looks great and should go through nicely my comment, on the use of `unconstrained` is not blocking.

PR #164, opened 2020-11-16
Downgrading pytorch from 1.7.0 to 1.6.0 for prod envs

dotsdl:
Latest pytorch can break e.g. psi4 due to requiring a substitute `dataclasses` package for the standard lib one. Using 1.6.0 doesn't do this.


trevorgokey (approved):
Fingers crossed!

PR #165, opened 2020-11-16
Added defunct, himem filtering/tagging to error cycling

dotsdl:
This PR Introduces filtering of optimizations in our error cycling lifecycle step in order to reduce compute waste. This additional error cycling logic has been introduced for both OptimizationDatasets and TorsionDriveDatasets.

It does the following:
1. For optimizations whose errors are due to SCF convergence, geometry optimization convergence, or basis coverage (with ANIxx currently), these are marked with compute tag `openff-defunct` and no longer receive restarts.
2. For optimizations whose errors appear due to high memory usage, these are marked with compute tag `openff-himem` and continue to receive restarts.
3. For all other optimizations that errored for other reasons, restarts proceed as before.

The error cycling report now also includes a section on high memory error counts. This information is actionable, since a rising count indicates that more high memory workers need to be provisioned to handle these identified high-memory tasks.

Compute tags allow for routing of tasks to workers with the matching compute tag. In the case of `openff-defunct`, this amounts to a dead-letter; we generally will not run workers with this compute tag. For the case `openff-himem`, only workers with memory per task > 48GiB should use this compute tag.


management/management.py
```diff
@@ -166,6 +166,53 @@ def count_unique_optimization_error_messages(
...
+
+def filter_scf_convergence(optimizations, return_filtered_instead=False):
+
+    to_match = "Could not converge SCF iterations"
+
+    return filter_optimizations(optimizations,
+                                to_match,
+                                return_filtered_instead=return_filtered_instead)
+
+def filter_opt_convergence(optimizations, return_filtered_instead=False):
```
jthorton, line 191: This one is a little awkward, I agree we do see a lot of these but I would say that most of the time they do vanish if you look through the error cycle reports on the past few datasets we see they gradually resolve as they are restarted, so maybe it would be best to hold off on this one.

dotsdl: Thanks @jthorton, good point! I've commented out this filter's usage in the list.

PR #169, opened 2020-12-01
(WIP) Restrained opt set with b3lyp-nl, wb97x-v/def2-tzvpd and mp2/heavy-aug-cc-pvtz for theory bm +  added 2020-12-14-theory-bm-torsiondrive-set-v1.0

hyejang:
- [ ] Ready to submit!


jthorton:
@hyejang following on from yesterday I am going to split up this PR slightly to keep track of whats going on. Ill open a separate PR for the torsiondrive benchmark and we can get that going today and then we can look at combining the restrained optimization sets in the same way. I also noticed that a molecule and 24 optimizations are missing from the set which I don't think is intentional.

hyejang:
@jthorton yea, there was one torsion scan used to be an ERROR job in so I could not generate 24 restrained optimizations from it. But since now it is on COMPLETE state, once i regenerate the dataset, we won't have any missing optimizations.

PR #170, opened 2020-12-01
Update to the new qcsubmit conda package.

jthorton:
Update to the new qcsubmit conda package which should fix errors when we try to add specifications which are already present in a dataset.


jthorton:
Looks like this is now pulling in the correct version of qcsubmit.

PR #176, opened 2021-01-19
Updates to prod-envs, GHA conda envs

dotsdl:
Updated the following packages across all environments:
- qcsubmit=0.1.1 -> openff-qcsubmit=0.2.0
- qcfractal=0.15.1 -> 0.15.2
- openforcefield=0.8.0 -> 0.8.3
- openmm=7.4.2 -> 7.5.0

Added to prod envs with psi4:
- gau2grid=2.0.3

Leaving in place, despite possible package updates (discuss):
- ambertools=20.4
- psi4=1.4a2.dev1058+670a850

### DIFF
--- a/devtools/conda-envs/error-cycle.yaml
+++ b/devtools/conda-envs/error-cycle.yaml
@@ -10,7 +10,7 @@ dependencies:
   - python
   - pip
   - qcportal
-  - qcsubmit==0.1.1
+  - openff-qcsubmit =0.2.0
   - pip:
       - PyGithub
       - pandas
--- a/devtools/conda-envs/queued-submit.yaml
+++ b/devtools/conda-envs/queued-submit.yaml
@@ -11,7 +11,7 @@ dependencies:
   - python
   - pip
   - qcportal
-  - qcsubmit==0.1.1
+  - openff-qcsubmit =0.2.0
   - pip:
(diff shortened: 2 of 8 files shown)


dotsdl:
I've tagged both @trevorgokey and @jthorton as reviewers. If one of you has a moment, let me know if you think these pin updates look reasonable. I'm testing each prod env now to make sure they install successfully on Linux.

dotsdl:
Confirmed: all prod envs can be installed on Linux without dependency resolution issues.

jthorton (approved):
LGTM, the only change I can think of with PSI4 would be the ability to now store all of the wavefunction in the schema in the latest versions.

dotsdl:
Awesome, thank you @jthorton! Will await your evaluation of latest Linux packages for [psi4](https://anaconda.org/psi4/psi4/files), [gau2grid](https://anaconda.org/psi4/gau2grid/files). Once we have confirmation from you that these behave as we expect, we'll add them to the envs here and deploy!

jthorton:
@dotsdl  I was able to get all wavefunction properties using `gau2grid=2.0.3` and `psi4=1.4a3.dev26+51d7763` I would recommend these pins going forward.

dotsdl:
Awesome, thank you so much @jthorton! Launching!

PR #178, opened 2021-01-26
[pepconf-dlc] Added default-dlc compute spec to PEPCONF dataset

dotsdl:
- [ ] Ready to submit!

### DIFF
--- a/submissions/2020-10-26-PEPCONF-Optimization/README.md
+++ b/submissions/2020-10-26-PEPCONF-Optimization/README.md
@@ -57,3 +57,10 @@ Generating stereoisomers, conformers, and QCSubmit dataset:
         - method: WB97X
         - basis: 6-31G*
         - program: psi4
+    - default-dlc
+        - method: B3LYP-D3BJ
+        - basis: DZVP
+        - program: psi4
+        - optimization_procedure:
+            - coordsys: dlc
+            - reset: true
--- a/submissions/2020-10-26-PEPCONF-Optimization/compute-dlc.json
+++ b/submissions/2020-10-26-PEPCONF-Optimization/compute-dlc.json
@@ -1,0 +1,64 @@
+{
+  "qc_specifications": {
+    "default-dlc": {
+      "method": "b3lyp-d3bj",
+      "basis": "dzvp",
+      "program": "psi4",
+      "spec_name": "default-dlc",
+      "spec_description": "Standard OpenFF optimization quantum chemistry specification, with DLC internal coordinates.",
+      "store_wavefunction": "none",
+      "implicit_solvent": null
+    }
+  },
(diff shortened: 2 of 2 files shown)


dotsdl:
@pavankum please review my parameter choices, and let me know if I need to change or add anything for this submission! This is the re-roll of PEPCONF with DLC internal coordinates and `reset=True` for geomeTRIC. 

I've also upped the `maxiter` for SCF to 300 and for geomeTRIC to 500 as a bit of hedging in case some of these crazy molecules still require quite a bit to converge.

pavankum (approved):

pavankum:
@dotsdl this looks great, I don't have anything else to add to the parameter choices now, you can merge.

dotsdl:
@pavankum thank you! Merging!

dotsdl:
Hmmm...something appears wrong with this submission. We get a response of 0 from QCArchive, and when I do a
```
from qcportal import FractalClient

cl = FractalClient()
ds = cl.get_collection("OptimizationDataset", "OpenFF PEPCONF OptimizationDataset v1.0")
ds.status()
```
I get:
```
            default  ani2x-qm  default-dlc
COMPLETE       1805       NaN          NaN
ERROR          5749    7560.0          NaN
INCOMPLETE        6       NaN          NaN
```

@jthorton do we know this could be an issue in latest QCSubmit?

dotsdl:
Also, question before we proceed much further: do we want wavefunction data saved for these optimizations?

jthorton:
Just had a little look into this and doing `ds.data.history` shows all the specs that have been used to create tasks and I get 
>  {'ani2x-qm', 'default'}
I think this is due to the threading update in QCSubmit, the tasks are chunked up between threads using the list of submitted indices, as this is a compute expansion there are no new indices so no tasks are made. Ill make a fix in QCSubmit to create tasks based on the entries present in the dataset if we are not adding any new entries which should fix this. 

On the wavefunction side, I am not sure if we tested if saving during optimizations would save it for every result in the trajectory which might be a lot of data which might never be used (not 100% sure) we can always calculate it for the optimised geometries if we need after.

dotsdl:
Excellent, thank you @jthorton! We'll try submitting again once we have a fix in QCSubmit, no problem.

Re: wavefunctions, agreed. We'll leave these off since that could be a lot of data that no one is interested in being stored. If it can be done in post, that's perfect.

dotsdl:
@jthorton is putting together a fix in `openff.qcsubmit` to allow resubmission of this PR; aiming to resubmit today.

dotsdl:
I've set this up to Error Cycle again; however, compute tag on all tasks has been set to `openff-himem`, which will restrict execution to only managers with this tag included in their supported tag list. This is currently the case for our PRP deployment.

PR #184, opened 2021-02-12
Added qcarchive-user-submit production environment

dotsdl:
Moved @pavankum's user submission environment to `prod-envs`, denoting it as a production resource. This is designed as a standard starting environment for preparing submissions, which we maintain to keep pins up-to-date with current practice and packages.


dotsdl:
@trevorgokey or @jthorton, a review suggesting improvements or changes from either of you and we can merge. We can then ask users to:

```
conda env create openforcefield/qcarchive-user-submit
```

going forward when preparing their submissions if they encounter issues. Thanks!

PR #173, opened 2020-12-18
TorsionDrive benchmark update

jthorton:
- [X] Created a new folder in the submissions directory containing the dataset
- [X] Added README.md describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [X] All files used to produce the dataset are included with a description
- [X] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [X] Dataset is named `dataset.json`
- [X] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Ready to submit!

## Info
This dataset combines the theory benchmark torsiondrives made by @hyejang including the addition of some new charged molecules into a single dataset with multiple specs and will replace all theory benchmark sets. The inputs have been reused from past datasets which should ensure task deduplication.

PR #185, opened 2021-02-16
Constrained optimization dataset for benchmarking (separated from PR #169)

hyejang:
- [x] Ready to submit!
#### Description
 This data is from PR #169, which contains (1) torsiondrive dataset and (2) constrained optimization datasets for theory benchmarking. This PR separates the optimization datasets from the previous PR for validation and submission. (The torsiondrive dataset was separated into PR #173.)


hyejang:
@dotsdl Once it is merged to master and submitted to QCA, I believe we are good to remove the previous PR #169. Does this sound right?

dotsdl:
Thanks @hyejang! Two questions:

1. Can we call these "OpenFF Theory Benchmarking Optimization Set ..."? I think it would be valuable to distinguish them from the TorsionDrive datasets of similar name. I can make this change throughout the files.
2.  I was going to suggest we merge these three datasets into one, but I noticed the `mp2-heavy-aug-cc-pvtz` set has the keywords for each entry, but the other two do not:
```
      "keywords": {
        "scf_type": "df",
        "mp2_type": "df",
        "guess": "sad"
      }
```
  Is this intentional? If so, I don't think merging them into one dataset would work, since this would apply to all specs. We can proceed with these three separate datasets, but I'm interested in @jthorton's thoughts on this.

hyejang:
1. I am not sure which name sounds more intuitive. To me, `OpenFF Theory Benchmarking Optimization Set ...` sounds like a dataset with fully optimized geometries, and considering the datasets have constrained optimizations along several torsion scans, if we want to add `Optimization` in the dataset name, `OpenFF Theory Benchmarking Set(Constrained Optimization) ...` or `OpenFF Theory Benchmarking Constrained Optimization Set ...` may work, although they seem too long. 
2. yea since the MP2 calculations need additional options, I added the psi4 options in keyword. If there's proper way to add additional input options, please let me know!

dotsdl:
@hyejang thank you!

1. Ah, I see your point here. I don't mind too long if you don't; I'd rather err on the side of descriptive in the name if we can. I can make the change to include "Constrained Optimization" in the names.
2. I think this may be the way to do this, so perfectly happy with this approach. I notice also that each spec has the name `default`, which is non-ideal but is consistent with the other submissions of this flavor, so I'm in favor of keeping it. If we find a way to roll these together into one dataset as we did for #173, then we can give each spec a less ambiguous name for that one.

dotsdl (approved):
Thank you @hyejang! Should be good to go. About to merge.

jthorton:
> I was going to suggest we merge these three datasets into one, but I noticed the mp2-heavy-aug-cc-pvtz set has the keywords for each entry, but the other two do not:
      "keywords": {
        "scf_type": "df",
        "mp2_type": "df",
        "guess": "sad"
      }
Is this intentional? If so, I don't think merging them into one dataset would work, since this would apply to all specs. We can proceed with these three separate datasets, but I'm interested in @jthorton's thoughts on this.

Ah yeah this is awkward and could easily be one dataset if we update qcsubmit to have keywords on a per spec basis lets aim to get this fixed so we can condense these back down in future.

jthorton:
@hyejang @SimonBoothroyd I just noticed that the WB97X-V optimisations are showing a much higher failure rate than the other methods I wonder if we need different settings for this functional, looking at the paper [here](https://pubs.rsc.org/en/content/articlelanding/2014/CP/c3cp54374a#!divAbstract) I see that it was optimised using a finer integration grid than the default in psi4 and also with a different basis set.

dotsdl:
Switching off error cycling on these datasets for now by removing `tracking` label. It's possible that additional forward progress can be gained on the largely incomplete `OpenFF Theory Benchmarking Constrained Optimization Set WB97X-V def2-TZVPD v1.0` if we allow the first errors to roll through without being restarted.

pavankum:
WB97X-V takes a lot of time due to VV10 dispersion being calculated using finite differences, I think this spec is of no use to us in the benchmarking work unless we see any improvements on psi4 side with the analytical gradients. Same with B3LYP-NL, nonlocal dispersion has no analytic gradients :( . I think we can stop and archive these or they would still be valuable, any thoughts @hyejang @jthorton ?
cc: @dotsdl

dotsdl:
If we would like to avoid computing these, then I can set them to the `openff-defunct` compute tag. @hyejang, what do you think?

dotsdl:
Setting to low priority until we have resolution on whether to continue with this submission.

pavankum:
Thank you @dotsdl , I think Hyesu is busy with dissertation and will be slow to respond for a few days.

PR #195, opened 2021-03-19
OpenFF | XTALPI : Industry Benchmark Season 1 - QM

dotsdl:
Public industry dataset from XTALPI for OpenFF geometry optimization benchmarking.

This dataset corresponds to the "Season 1 - QM" stage,
and was prepared using the steps given in the [season 1 protocol](https://openforcefield.atlassian.net/wiki/spaces/PS/pages/971898891/Optimization+Benchmarking+Protocol+-+Season+1).

The following software versions were used for preparation:
`openff-benchmark` 2021.3.17.0
`openforcefield` 0.8.4
`openff-qcsubmit` 0.2.1
`rdkit` 2020.09.3

- [ ] Ready to submit!


jthorton:
@dotsdl Thanks for starting to put these together, I am slightly confused as to how B and Si got through the coverage report from openff-benchmark as they should be removed? Another thought is that if these datasets are to be shared with the publication would we want to put them all together in one dataset to make finding them easier? We could still track the provenance of each molecule by adding the company tag in the entry attributes so they can be separated later. This might also reduce the amount of datasets created by this project as each one will have an accompanying mm dataset.

PR #196, opened 2021-03-24
OpenFF Amide Torsion Set v1.0

SimonBoothroyd:
- [x] Make a new dataset entry in the mapping table, INDEX.md
- [ ] Ready to submit!


trevorgokey (approved):
Looks good! It could be submitted as-is, but I have one suggestion.

Using only one conformation can give situations where SCF or optimization can fail to converge on that one molecule conformation (from the MM-based toolkit), killing the entire TD. To get around this, I'd jump the number of conformers up so that there is a better chance that at least one will optimize successfully. The TD will then take the successful optimizations, and proceed with the wavefront. This was the reasoning as to why I used 10 confs with an RMS of 0.1 in the WBO set.

The fact that the mean confs is 1.26 tells me that a single molecule has two TDs, and qcsubmit has combined them into a single molecule object, even though there will be two separate TD services.

jthorton (approved):
I agree with @trevorgokey I think expanding conformers would be wise here there is a [component](https://github.com/openforcefield/openff-qcsubmit/blob/45e716e3cc70ee05614fbe951c627efc6d9eec84/openff/qcsubmit/workflow_components/conformer_generation.py#L15) to do this in parallel. Note that due to how deduplication works the conformers produced may be different from what you have generated here.

dotsdl (approved):
Fantastic @SimonBoothroyd! This is a clean, clear submission. Thanks for putting it together.

SimonBoothroyd:
@jthorton @trevorgokey @dotsdl these still seem to be erroring - can one you take a quick look when you get a chance?

trevorgokey:
There were still 581 optimization errors after a manual GHA, so I just restarted this dataset using local tools. It looks like the GHA is missing some optimizations to restart.

SimonBoothroyd:
Thanks @dotsdl - I think what's finished is fine so maybe just move to "Archived/Complete"

PR #199, opened 2021-03-30
first iteration to change README.md

pavankum:
just concatenated lifecycle and index files


jthorton (approved):
Great idea this has been needed for a while.

PR #197, opened 2021-03-29
Add OpenFF Aniline 2D Impropers v1.0

SimonBoothroyd:
- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added README.md describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset is named `dataset.json`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [ ] Make a new dataset entry in the mapping table, INDEX.md
- [x] Ready to submit!

PR #200, opened 2021-04-02
Add OpenFF Aniline Para Opt v1.0

SimonBoothroyd:
- [ ] Ready to submit!


trevorgokey (approved):
This looks good. My only comment is that we should be letting qcsubmit load the smi so that it can record it as a filtered molecule. However, it's not worth changing anything at the moment, so I'll submit the notebook as is.

dotsdl:
@SimonBoothroyd should we up the priority to *high* for this one, since it blocks #197?

PR #204, opened 2021-05-07
Add expanded Gen3 torsion set

hyejang:
## New Submission Checklist

- [ ] Ready to submit!


hyejang:
@trevorgokey I remember you mentioned that we'd want to add `changelog` for the expanded dataset, but I found that we do not have `changelog` field in the qcsubmit Metadata class. Do we want to add the new field, or is there any other place that i am supposed to add `changelog`?

dotsdl (approved):
Looks fantastic @hyejang! This is a work of art! I've fixed the validation to handle `dataset*.json*`, so we should get a validation report.

@SimonBoothroyd due to the size of this PR, and my relative unfamiliarity with the methodology used to construct the dataset, can you give this a review as well? Once satisfied, we should be clear to fire away!

trevorgokey:
Hi @hyejang apology for the delay. I don't think there is a way to add a changelog to the dataset yet via qcsubmit; that will work once the standards are implemented. I think for now, given the intended use of this dataset, is to just write the changelog to this github repo readme. It will be useful when the final, citable dataset is generated, and we can incorporate any new info from the changelog into the final dataset description. Thanks!!

PR #211, opened 2021-06-22
OpenFF BCC Refit Study COH v2.0

SimonBoothroyd:
## New Submission Checklist

- [ ] Ready to submit!

- [ ] Ready to submit!


SimonBoothroyd:
@trevorgokey @jthorton @dotsdl this should be ready for review

trevorgokey:
This generally looks good, but there are some yellow flags and it would be great to have some clarification.
1. I am not sure if naming this v2.0 makes sense given that there is no v1.0.  I think it is important to note that the version specifies the version of the dataset type, not the "study" per se. As a reviewer I know there is a `Dataset` v1.0, but this is wildly opaque to users. This inconsistency also has some long-term downsides that I think need to be addressed. Since the grid kwargs make the calculations incompatible with previous specs, I was trying to figure out if new spec names were used with the new grid since this is v2.0 and v1.0 should be somewhere. Knowing that v1.0 is a `Dataset`, I see the compute specs use the same name as v2.0 with different grid. This can potentially ruin a user's science if these two datasets are cross-referenced. It is inappropriate to compare energies of the same model with different grids.
2. I wish a changelog would be somewhere so I could have checked this without opening up a `FractalClient` and digging, especially since the notion of a v2.0 assumes a v1.0 somewhere.

I think some changes are needed, but I think there are a few ways to solve this. I will list the 2 immediate ones that come to mind.
1. One solution is to use a new dataset name from the `Dataset` name and call this one v1.0; there would be less cohesion, but would be clear, and safe from subtle bugs introduced by using the same dataset name, same spec name, but different settings.
2. Rename the v2.0 specs to reflect the grid changes and to distinguish them from the v1.0 specs.
2.1. Add a v2.0 `Dataset` with the same new spec names, so that the v1.0 `Dataset` geometries/energies can be cross-compared with these new optimized geometries/energies.

dotsdl:
@SimonBoothroyd did the discussion earlier today address your questions sufficiently for this submission? If so I'll review and get it going.

SimonBoothroyd:
> @SimonBoothroyd did the discussion earlier today address your questions sufficiently for this submission? If so I'll review and get it going.

Yeah this should be G2G!

dotsdl (approved):
Excellent @SimonBoothroyd! Thank you for putting this together!

dotsdl:
Getting fairly high error rates on remaining optimizations. Switching to `priority-normal` to allow some other submissions time to execute.

dotsdl:
Is `pw6b95/aug-cc-pV(D+d)Z` known to not use all the threads given to `psi4` via QCEngine? On PRP I'm seeing low resource usage for containers running jobs with this compute spec.

dotsdl:
Set priority high from discussion today with @SimonBoothroyd; we will see if the SCF convergence errors of PCM case advance in the next week. After that, could try routing to dedicated infrastructure with compute tag to see if SCF convergence errors are truly systematic.

@jthorton feel free to drop any conclusions from your discussions with @SimonBoothroyd here.

dotsdl:
Dropped `tracking` tag to remove from error cycling to avoid restarts of PCM runs, which are expensive and appear to be clogging our QM compute. Will restart errored non-PCM cases manually, and run these on dedicated compute infrastructure on PRP.

dotsdl:
Performed a few cycles of the following:

```
> ds = fc.get_collection('OptimizationDataset', "OpenFF BCC Refit Study COH v2.0")
> ds.status()
            hf/6-31G*  pw6b95/aug-cc-pV(D+d)Z/uf  pw6b95/aug-cc-pV(D+d)Z/uf,pcm
COMPLETE       1550.0                       1365                             57
ERROR             NaN                        177                            194
INCOMPLETE        NaN                          8                           1299

> opts = [opt.id for opt in ds.df['pw6b95/aug-cc-pV(D+d)Z/uf'].values]
> fc.modify_tasks(operation='restart', base_result=opts)
```

Seeing consistent errors among the remaining `pw6b95/aug-cc-pV(D+d)Z/uf ` entries.

@SimonBoothroyd, can we move this submission to "End-of-Life"? I have allocated compute resources devoted to those remaining entries elsewhere now.

PR #215, opened 2021-07-02
Update constrained opts of mp2-heavy-aug-cc-pvtz

pavankum:
## Updated Submission Checklist

- [x] Updated `README.md` describing the change 
- [x] Ready to submit!


pavankum:
looks like even the MP2 spec data is useless since the constrained opts are not aligned properly with the torsion scans, this may be discarded, will keep it on hold for now, no need to review this one yet

dotsdl:
No problem @pavankum. Happy to review when you've got a dataset you'd like submitted!

pavankum:
> looks like even the MP2 spec data is useless since the constrained opts are not aligned properly with the torsion scans, this may be discarded, will keep it on hold for now, no need to review this one yet

Note to self: the mistake with misalignment is due to bad angle index matching, somehow the constrained structures correspond to different ones after the creation of new dataset, easy way to find them is to measure the dihedral angle from the geometries.

pavankum:
Hi @dotsdl , when you have time can you review and push this through. This includes addition of 23 x 24 new constrained opts on top of the current 36 x 24 set. These will be faster and do not consume much time since they're single optimizations. I had rerun the notebook with few updates to the qcsubmit invocations, so the dataset file created includes the previous constrained opts that are already calculated, I am assuming these would be filtered out at the QCA server level as duplicates (correct me if I am wrong).
This is supposed to be an alternative to the full mp2 torsion scans I submitted in #217 , the other set might become redundant but I still want to confirm a few cases that show completely different torsion profiles so I am still inclined to let them run.

dotsdl:
Can do! Probably can get this in tomorrow morning!

dotsdl:
@pavankum we should set the version in the dataset name to `v1.1`, since this amounts to an expansion of the previous set. Can you also put a note on what the `v1.1` version includes, expanding on the `v1.0` version?

pavankum:
@dotsdl Thank you for the review! I just pushed those changes, one quick question, if I remade the dataset.json with additional molecules the old calculations would be marked as duplicate at the server level and won't be calculated again, right?

dotsdl:
> @dotsdl Thank you for the review! I just pushed those changes, one quick question, if I remade the dataset.json with additional molecules the old calculations would be marked as duplicate at the server level and won't be calculated again, right?

Should be, are they completely unchanged? 

I'm going to to put the original `dataset.json` in place and name the new one `dataset-v1.1.json`. I'm also going to rename the original `QCSubmit_workflow_v1.ipynb` back to `QCSubmit_workflow.ipynb`, and ditto for the other files. In the future I think we should avoid renaming files from previous submissions where possible, and only add files when making new iterations.

dotsdl:
@pavankum is the `torsiondrive_data.pickle` necessary? If not, I'll remove it!

pavankum:
@dotsdl yeah, they're unchanged, I hope the change in qcsubmit version wouldn't introduce anything new in those records. 

> we should avoid renaming files from previous submissions where possible, and only add files when making new iterations.

Sure, I will keep that in mind for future submissions, sorry about that.

> is the torsiondrive_data.pickle necessary? If not, I'll remove it!

I think we can remove that, we are not reading it again, I just used the old submission script.

dotsdl:
Awesome, thanks @pavankum! Finishing up my review. The diff is a lot easier to view now! Thank you!

pavankum:
Thank you @dotsdl :)

dotsdl (approved):
Looking good @pavankum! Sorry for the delay; submitting!

PR #219, opened 2021-08-10
Updates to all production environments given new packages

dotsdl:
In particular, psi4 1.4 is out.
Also want to use new namespace packages for toolkit, others.

### DIFF
--- a/.github/workflows/prod-envs.yml
+++ b/.github/workflows/prod-envs.yml
@@ -14,6 +14,16 @@ on:
 jobs:
   deploy-conda-env:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        env-name:
+          - qcarchive-user-submit
+          - qcarchive-worker-openff
+          - qcarchive-worker-openff-ani
+          - qcarchive-worker-openff-openmm
+          - qcarchive-worker-openff-psi4
+          - qcarchive-worker-openff-xtb
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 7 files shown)


dotsdl:
@pavankum can you try out the `qcarchive-worker-openff` and `qcarchive-worker-openff-psi4` envs from this branch for some local `psi4` executions? 

I'm going to add a GHA for some ad-hoc `psi4` tests to this PR next for a few molecules and procedures, but some independent validation would help make sure this combo of packages works well for `psi4` before we deploy at scale.

pavankum:
Great! Sure @dotsdl, I will do some local tests with psi4 envs. Btw since Sage will be out in the next two days we can hold on the mm updates(?)

dotsdl:
Yes, we can wait for the Sage release if it's bound to be out later this week. No problem!

j-wags:
Just chiming in: I wouldn't hold off just for the full Sage release. It's _anticipated_ in the next few days, but the purpose of the `rc` system is to be uncertain and keep the window open for another month of delay if problems are found. So I think this PR should go ahead with the `rc.2` package.

jthorton:
`qcarchive-worker-openff-psi4` builds fine and runs gas-phase calculations, I did hit a [bug](https://github.com/psi4/psi4/issues/2272) with an scf property namely `MBIS_CHARGES` but I don't think we are interested in these for now.

pavankum (commented):
@dotsdl one minor change to `qcarchive-worker-openff.yaml` as well as `qcarchive-worker-openff-openmm.yaml`, it needs the channel `conda-forge/label/openff-forcefields_dev` for Sage-rc2. Other than that everything else looks good to me. I tested the psi4 env too locally.

dotsdl:
Awesome, thanks all! Getting ready to merge this and test deployments out!

pavankum:
Thank you @dotsdl! You can merge it.

PR #221, opened 2021-08-13
Added initial stab at psi4 prod testing

dotsdl:
We want to add our own tests for known problem molecules/procedures as we identify
them, as well as run the `openff-qcsubmit` test suite for prod envs
containing psi4.


dotsdl:
We might need to merge a prototype in order to actually test out if the Action is working or not. :/

dotsdl:
Looks like we have some work to do on `openff-qcsubmit` with regard to latest OpenMM. Otherwise also seeing sqlalchemy errors, which look like issues in interacting with a local QCFractal server.

dotsdl:
Looks like the OpenMM issues on `openff-qcsubmit` will be addressed in the next release. Will hold on this until we have that.

dotsdl:
@jthorton any advice on getting past the `openff-qcsubmit` errors we see here? Not clear if we're doing something fundamentally different than the `openff-qcsubmit` CI to cause these errors.

jthorton:
@dotsdl I think this might be to do with the environment being incomplete compared to the qcsubmit testing environment. The torsiondrive errors for example are probably due to torsiondrive not being installed.

dotsdl:
Test failures probably probably due to tasks staying in INCOMPLETE forever because the manager doesn't have the software to execute the submissions. Will pursue marking tests in `openff-qcsubmit` in a way that allows us to disable on a `pytest` CLI call.

PR #222, opened 2021-08-20
Update prod envs to use openff-forcefields 2.0.0

dotsdl:
`openff-forcefields` 2.0.0 was released this week. This PR adds production env support for use of Sage in MM computations.


dotsdl:
I'm going to go ahead and merge. This replaces our use of the `rc.2` version of the same package, so is a small incremental change.

PR #223, opened 2021-08-25
Add version 2 of the benchmark ligand fragments

jthorton:
## New Submission Checklist
- [X] Created a new folder in the submissions directory containing the dataset
- [X] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [X] All files used to produce the dataset are included with a description
- [X] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [X] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [X] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [x] Ready to submit!

# Info
This PR updates the `OpenFF-benchmark-ligand-fragments-v1.0` to version 2 by supplementing the dataset with new fragments produced using the reworked fragmenter with both the openeye and ambertools back ends.

PR #220, opened 2021-08-12
add single point energy calculations

pavankum:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [x] Ready to submit!

PR #224, opened 2021-08-25
Validation update

jthorton:
## Management update
This PR reworks some of the current validation workflow to work with the dataset update in the newest version of qcsubmit.

### DIFF
--- a/management/validation.py
+++ b/management/validation.py
@@ -188,6 +188,8 @@ def check_qcspec_coverage(dataset_data):
         valid_scf_props = check_scf_props(spec)
         try:
             # remove the scf props so they dont cause issues
+            # we already validated them explicitly above in the call to `check_scf_props`
+            # not necessary to validate them twice, and we want to distinguish them from other errors
             del spec["scf_properties"]
             dataset.add_qc_spec(**spec)
             validated = check_mark
--- a/management/validation.py
+++ b/management/validation.py
@@ -14,6 +14,7 @@ from github import Github
 from openff.qcsubmit.datasets import (BasicDataset, OptimizationDataset,
                                       TorsiondriveDataset,
                                       update_specification_and_metadata)
+from openff.qcsubmit.common_structures import SCFProperties, Metadata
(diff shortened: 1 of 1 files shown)


management/validation.py
```diff
@@ -179,14 +185,16 @@ def check_qcspec_coverage(dataset_data):
     # now try and add each spec
     spec_report = {}
     for spec in qc_specs.values():
-
+        valid_scf_props = check_scf_props(spec)
         try:
+            # remove the scf props so they dont cause issues
+            del spec["scf_properties"]
```
dotsdl, line 191: What kind of issues? I don't understand why we would delete the `"scf_properties"` key from the spec dict here.

dotsdl (approved):
Thanks @jthorton, sorry for the delay here. I have one item I don't understand. Can you clarify?

management/validation.py
pavankum: just curious about this, I am hitting validation errors on my new submission #220

management/validation.py
```diff
@@ -179,14 +185,16 @@ def check_qcspec_coverage(dataset_data):
     # now try and add each spec
     spec_report = {}
     for spec in qc_specs.values():
-
+        valid_scf_props = check_scf_props(spec)
```
jthorton, line 188: @dotsdl  @pavankum We now validate the scf properties separately to the whole QC specification in this line.

management/validation.py
jthorton: My thinking was as the scf properties are now validated separately on line 188 if they are incorrect they should not then raise an error when validating the method, basis and program so I remove them. The current errors in PRs #220 and #223 are due to the qcsubmit dataset changes which moved the `scf_properties` from the dataset to the qcspecifications.

dotsdl: Ah, understood. Thank you! I'll add these details to the inline comment.

dotsdl:
Merged! Thanks @jthorton! Apologies again for the delay on my part.

PR #234, opened 2021-10-11
prod envs: qcengine 0.19.0 -> 0.20.0

dotsdl:


jthorton:
@dotsdl qcelmental must also be `>=0.23.0` if we go for qcengine 20 see [here](https://github.com/MolSSI/QCEngine/blob/0612e81a2ce6685866ea40692fa95649b06e2b35/setup.py#L31)

dotsdl:
This is ready for review now, since we've now got QCFractal 0.15.7, which should be compatible with both QCEngine 0.20 and QCElemental 0.23.

@jthorton do you want to give this another pass and see if I missed anything?

jthorton:
@dotsdl Thanks for updating this! LGTM!

dotsdl:
Thanks @jthorton! Just tested all env installs here as well, and no conda errors. We'll proceed.

PR #239, opened 2021-11-09
ml_solvated_amino_acids submission files

pavankum:
## New Submission Checklist

- [x] Ready to submit!


pavankum:
@JoshHorton @dotsdl when you've a moment can you please take a look at the validation error with metadata

dotsdl (approved):
Looks great @pavankum! Can you adjust the name to be:

```
SPICE Solvated Amino Acids Single Points Dataset v1.0
```

That should be it!

pavankum:
@peastman we just pushed this small dataset to compute as a guinea pig, and if this goes through well without any further glitches we will merge all other submissions too :slightly_smiling_face:

peastman:
Excellent, thank you!

peastman:
Does the above report just mean it's still running, or did something go wrong?

dotsdl:
I'm working on it now @peastman. It might be getting outcompeted by older submissions.

dotsdl:
Oddly, there were no `ResultRecords` associated with the dataset. I just did a:

```
from qcportal import FractalClient
fc = FractalClient.from_file()

ds = fc.get_collection('Dataset', 'SPICE Solvated Amino Acids Single Points Dataset v1.0')
ds.compute(method='wb97m-d3bj')
```

to generate them. Prioritized all to `high` and added compute tag `openff`. Unclear why these result records didn't exist, as that is part of the submission process.

peastman:
What do you mean by "a continuous order"?

peastman:
The atom ordering was chosen by OpenFF Toolkit:

```python
mol = mol.canonical_order_atoms()
smiles = mol.to_smiles(isomeric=True, explicit_hydrogens=True, mapped=True)
```

It appears to put all heavy atoms before all hydrogens, regardless of molecule.  For example, here is the SMILES string for solvated alanine.

```
[O:10]([H:45])[H:46].[O:11]([H:47])[H:48].[O:12]([H:49])[H:50].[O:13]([H:51])[H:52].[O:14]([H:53])[H:54].[O:15]([H:55])[H:56].[O:16]([H:57])[H:58].[O:17]([H:59])[H:60].[O:18]([H:61])[H:62].[O:19]([H:63])[H:64].[O:1]=[C:23]([N:26]([C@@:30]([C:24](=[O:2])[N:25]([C:28]([H:76])([H:77])[H:78])[H:71])([C:29]([H:79])([H:80])[H:81])[H:82])[H:72])[C:27]([H:73])([H:74])[H:75].[O:20]([H:65])[H:66].[O:21]([H:67])[H:68].[O:22]([H:69])[H:70].[O:3]([H:31])[H:32].[O:4]([H:33])[H:34].[O:5]([H:35])[H:36].[O:6]([H:37])[H:38].[O:7]([H:39])[H:40].[O:8]([H:41])[H:42].[O:9]([H:43])[H:44]
```

jthorton:
@peastman Good point I think this is due to the use of `canonical_order_atoms`  which is mixing the different molecules together. I will try this again without this locally and report back.

jthorton:
Removing tracking and moving to review as the failure is systematic.

dotsdl:
Thanks @jthorton! You beat me to it; completely agree.

jthorton:
@peastman I have confirmed this locally, if I remake the input without putting the atoms in canonical order the calculation works. There is also one fix needed in qcsubmit to ensure the indices of each fragment are ordered then we should be good to go.

peastman:
Great!  Do we need to regenerate the HDF5 file with different ordering, or can qcsubmit handle it?  If the former, how should I define the order?

jthorton:
I plan to make qcsubmit reorder the molecules to ensure they are correct but this will make the input inconsistent with the calculations on qcfractal. So it depends if this is an issue or not?

peastman:
The HDF5 files are just an intermediate step in creating the dataset, so I don't think that's a problem.  I assume the final dataset people download from QCArchive will contain a SMILES string, positions, and forces that all use a single consistent order?  I also assume it will include the human readable IDs from the HDF5 file, e.g. solvated alanine will be labeled as "ALA" and not just with a huge SMILES string?  That will let people link back to the corresponding record in the HDF5 file if they want to trace exactly how something was created.

PR #243, opened 2021-11-09
pubchem_set1

pavankum:
## New Submission Checklist

- [x] Ready to submit!


submissions/2021-11-08-QMDataset-pubchem-set1-single-points/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+ - Date: 2021.11.08
+ - Class: Basic dataset 
+ - Purpose: Energy calculation
+ - Collection: BasicDataset
+ - Name: SPICE PubChem Set 1 Single Points Dataset
+ - Number of unique molecules:        2372
+ - Number of filtered molecules:      0
+ - Number of conformers:              118606
+ - Number of conformers min mean max: 6  50.00 100
```
peastman, line 15: ??? If it has more than 50 conformers for any molecule, something is very strange!

pavankum: @dotsdl @jthorton any thoughts?

peastman: I ran the following test just to be sure.

```python
import h5py
f = h5py.File('pubchem-1-2500.hdf5')
for id in f:
  conf = f[id]['conformations']
  if conf.shape[0] > 50:
    print(id, conf.shape)
```

It doesn't report anything.

pavankum: @peastman Yeah, I checked the same. During dataset preparation qcsubmit does a deduplication step (based on inchi keys IIRC), and the output shows a drop from 2373 to 2372 molecules. I checked the inchi keys of the created molecules, and for these two molecules with smiles `'[O-][P-](=O)[O-]', '[O-]P([O-])[O-]'`, inchi key is same, `'AQSJGOWTSHOLKH-UHFFFAOYSA-N'`. Somewhat related to this issue https://github.com/openforcefield/openff-qcsubmit/issues/60

peastman: Is that going to cause problems?

pavankum: afaik it shouldn't matter since QM doesn't care about the 2D representation, and that one molecule record seems to have 100 conformers post dataset preparation

jthorton: Thanks for looking into this @pavankum I think you are right about the deduplication. Do we want to remove some of these conformers for this molecule so it only has 50 like the others.

jthorton:
I also noticed we have `I` in this dataset are we going to have an issue with density fitting like we had with dzvp?

submissions/2021-11-08-QMDataset-pubchem-set1-single-points/README.md
peastman: It's not a problem if there are more conformers.  It was just surprising.  And normally my filtering by ECFP4 fingerprint would have caught identical molecules, but this case is different because they really are the same as far as QM is concerned (same elements, same total charge), but the fingerprints are totally different.

pavankum:
> I also noticed we have `I` in this dataset are we going to have an issue with density fitting like we had with dzvp?

@jthorton Thanks for catching that, I just checked the auxiliary basis set used with def2-tzvppd, it is def2-universal-jkfit.gbs and it has elements up to Radon, so we're in safe zone.

submissions/2021-11-08-QMDataset-pubchem-set1-single-points/README.md
pavankum: @peastman Thank you for going through the PR and noticing this detail, marking this as resolved now.

jthorton (approved):
LTGM! There are some large charges in this dataset which I guess is due to the molecules being quite extreme?

submissions/2021-11-08-QMDataset-pubchem-set1-single-points/README.md
```diff
@@ -0,0 +1,50 @@
...
+Here any information regarding dataset changes are recorded.
+
+## QCSubmit generation pipeline
+
+ - `Dataset_Generation.ipynb`: This notebook shows how the dataset was prepared from the input files. 
+ 
+## QCSubmit Manifest
+
+- `Dataset_Generation.ipynb`: Dataset creation notebook with instructions for submission.
+- `dataset.json.bz2`: The compressed constrained optimization dataset ready for submission.
```
jthorton, line 32: These just need cleaning up a bit, they must have been copied from the last readme.

submissions/2021-11-08-QMDataset-pubchem-set1-single-points/README.md
```diff
@@ -0,0 +1,50 @@
...
+ - Date: 2021.11.08
+ - Class: Basic dataset 
+ - Purpose: Energy calculation
+ - Collection: BasicDataset
+ - Name: SPICE PubChem Set 1 Single Points Dataset v1.0
+ - Number of unique molecules:        2372
+ - Number of filtered molecules:      0
+ - Number of conformers:              118606
+ - Number of conformers min mean max: 6  50.00 100
+ - Set of charges: [-8.0, -4.0, -3.0, -2.0, -1.0, 0.0, 1.0, 2.0]
```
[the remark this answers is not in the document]
peastman: I didn't filter by charge, so there's no reason not to expect them.

dotsdl:
Thanks folks! Launching! :rocket:

dotsdl:
Attempting a call to `Dataset.compute` locally to push this through, troubleshoot the above error if I encounter it here.

jthorton:
@dotsdl I think this is probably due to the size of the dataset it is `118606` calculations, this may happen to all of the pubchem datasets.

dotsdl:
The local call to `Dataset.compute` went through. @jthorton yeah I'm curious if this ends up happening to every one of these. Not sure if the resource constraints on the Actions make a difference here.

dotsdl:
Called `Dataset.compute()` on this dataset manually with QCPortal 0.15.7.. Watching behavior.

pavankum:
@peastman : A heads-up on an unexpected bump, @jthorton pointed out a major bug in psi4 that [affects the range separated hybrid calculations](https://github.com/psi4/psi4/issues/2351#issuecomment-975326494), which is fixed right now in the dev branch, and we either have to modify the production envs after they make a release, or modify our submissions disabling the problematic `wcombine` option, and redo some of the finished sets. This may be done post thanksgiving vacation, next week, depending on @dotsdl and Josh's availability.

peastman:
Good to know, thanks.  Is that the cause of some of the failures it's been reporting?

pavankum:
Nope, the failures are mostly on the compute side. We submit to a bunch of pre-emptible workers and sometimes our jobs fail because of that, or some may run into out-of-memory issues, and the error cycling takes care of the failures. The tracebacks would give an idea of the errors and we usually poke in to check individual runs if the failure rate doesn't fall down over a period of two three days.

dotsdl:
Working to update prod envs now to use latest psi4 version 1.5.

pavankum:
revised set v1.1 is at #260 , this can be dumped/archived

PR #244, opened 2021-11-09
SPICE DES370K (multi-molecule)

pavankum:
## New Submission Checklist

- [x] Ready to submit!


pavankum:
@jthorton / @dotsdl another error to look into when you have time, 
```
ValidationError: 1 validation error for Molecule
connectivity
  ensure this value has at least 1 items (type=value_error.list.min_items; limit_value=1)
```

jthorton:
@pavankum this should be fixed in the newest toolkit release see the patch here https://github.com/openforcefield/openff-toolkit/pull/1101.

pavankum:
Thank you @jthorton :)

dotsdl:
@pavankum I see both `des370k.hdf5.tar.gz` and `des370k.tar.gz` in the directory. Is this intentional? It looks like they feature the same contents?

pavankum:
> @pavankum I see both `des370k.hdf5.tar.gz` and `des370k.tar.gz` in the directory. Is this intentional? It looks like they feature the same contents?

I might've included by mistake, will delete one

jthorton (changes requested):
Thanks for putting this together @pavankum. I think we need to confirm that the fragments in  the qcelemental molecules are in a continuous ordering before we submit or we will run into the error we see [here](https://github.com/openforcefield/qca-dataset-submission/pull/239#issuecomment-970322729)

pavankum:
> Thanks for putting this together @pavankum. I think we need to confirm that the fragments in the qcelemental molecules are in a continuous ordering before we submit or we will run into the error we see [here](https://github.com/openforcefield/qca-dataset-submission/pull/239#issuecomment-970322729)

@jthorton do you think those errors would be specific for the multi-molecule systems?

dotsdl:
@pavankum is this one ready to go too?

pavankum:
@dotsdl this is good to go, to be sure that there are no further issues may be we can push the smaller set of solvated amino acids  first #257 ?

dotsdl:
@pavankum I can review this one today too now that #257 is through; what do you think?

pavankum:
@dotsdl Sure, that'd be great!

dotsdl:
@peastman to confirm, you don't need wavefunctions on this dataset? I'd like to avoid it if possible, or hold off on submission if you are uncertain.

peastman:
Correct.  I don't need wavefunctions on any of the datasets.

dotsdl (approved):
This is ready to go; will likely need to be launched manually due to number of conformers (345k). I have set `store_wavefunction: "none"` for this set.

dotsdl:
@jthorton are you satisfied given [your comment earlier](https://github.com/openforcefield/qca-dataset-submission/pull/244#pullrequestreview-807470931)?

jthorton:
@dotsdl as long as this was regenerated with the fixed version of qcsubmit it should be fine.

dotsdl:
@pavankum and @bennybp, are either of you opposed to me submitting this? Or would this be playing with fire at this time?

pavankum:
@dotsdl I have no objections, feel free to merge if Ben agrees.

pavankum:
just for reference number of heavy atoms is less than 15 or so in each, so this shouldn't take up much space

dotsdl:
Cool, thanks! Merging and submitting! Also got an affirmative from @bennybp that we should be okay here.

dotsdl:
This submission is currently being submitted locally through `vulkan` due to its size.

pavankum:
since the error_cycling is choking a bit because of the larger dataset size I checked locally and the completion rate is impressive

|            |   spice |
|:-----------|--------:|
| COMPLETE   |  234068 |
| ERROR      |   43284 |
| INCOMPLETE |   68330 |
| NaN        |       0 |

peastman:
Thanks!  These are really small dimers.  The median size is 16 atoms, including hydrogens.  They go fast.

peastman:
I think PacificResearchPlatformQM is preventing us from completing this.  We could easily get through the remaining calculations in a single day.  But it grabs almost all of them and produces errors, which prevents any other cluster from doing them.

pavankum:
Yeah, it's a sticky situation, we opted for slow progress instead of zero progress on PRP until the new backend changes are adapted in the production environments. We can stop sending spice jobs to PRP by re-tagging this set and setting up managers specifically for that tag, @dotsdl any thoughts?

dotsdl:
I can also just shut off our PRP workers devoted to SPICE. I'll do that if that's preferable at this point.

dotsdl:
Deleted SPICE deployment on PRP.

peastman:
Thanks!  We can turn them back on once the fix is deployed.

pavankum:
All the errors are from MBIS charge calculation, except one with SCF convergence, if 99% completion rate is acceptable we can move this dataset to end of life so that we can avoid rerunning the consistent failures.

peastman:
It's probably acceptable.  Can we retrieve a list of which samples had the errors?

pavankum:
Here is a notebook to extract the failed inputs, I picked the smiles of first 100 errors as a sample

[jupyter_notebook.tar.gz](https://github.com/openforcefield/qca-dataset-submission/files/8052486/jupyter_notebook.tar.gz)

pavankum:
If the errors are only with MBIS charges we can build a supplementary dataset with those particular records, removing MBIS calculation.

peastman:
That would be great.  The charges are only there because we thought they might be useful to someone.  We're mostly interested in forces and energies though.

peastman:
It's interesting that all of the first 100 involve a Na+ or Li+ ion interacting with some larger, neutral molecule.  I don't know if there's a particular reason those tend to be more difficult to solve.

PR #252, opened 2021-11-11
SPICE Dipeptides Optimization Set

pavankum:
## New Submission Checklist

- [x] Ready to submit!


dotsdl (approved):
Looks great @pavankum!

Some questions before we go further:

1. Do we want just a single conformer/optimization per molecule? Depending on the downstream needs we may benefit from having several conformers, as these may each find their own, different local minima.
2. Do you expect to need the wavefunction from any of these? We currently don't get wavefunctions attached to *any* of the `ResultRecord`s that make up the optimization sequence. I have in my queue to draft a solution to this in QCFractal, but it isn't in place yet.

pavankum:
> Some questions before we go further: Do we want just a single conformer/optimization per molecule? Depending on the downstream needs we may benefit from having several conformers, as these may each find their own, different local minima.
@dotsdl that's a great catch, thanks for pointing that! Yeah, that makes sense, will modify the submission with max_confs of 50. 

>  Do you expect to need the wavefunction from any of these? We currently don't get wavefunctions attached to any of the ResultRecords that make up the optimization sequence. I have in my queue to draft a solution to this in QCFractal, but it isn't in place yet.
hmmm, I guess we can get away with energies/forces for this one, John wanted an optimization set to compare which conf. generation scheme is doing well.

pavankum:
@dotsdl Thanks for your review, I updated the submission, let me know if this looks good now, and if I can go ahead with the squash merge.

jthorton:
Looks great @pavankum just 2 points:
1. can we remove the `orbitals_and_eigenvalues` as they won't be saved anyway and it helps keep it clear what is to be expected in the dataset.
2. The other peptide dataset has around 50 diverse conformers per molecule, maybe we should try and generate a lot more conformers here to make the two comparable?

pavankum:
> Looks great @pavankum just 2 points:
> 
>     1. can we remove the `orbitals_and_eigenvalues` as they won't be saved anyway and it helps keep it clear what is to be expected in the dataset.
Thank you for the review @jthorton , sure will update.
> 
>     2. The other peptide dataset has around 50 diverse conformers per molecule, maybe we should try and generate a lot more conformers here to make the two comparable?
Initially I thought of adding 50 conformers but since this is an opt set with def2-tzvppd I backed out fearing this may take forever to complete with the larger basis,. For a fair comparison it makes sense to have same number of confs, will update.

jchodera:
@pavankum Should this dataset have been terminated now that the [SPICE Dipeptides revised v1.1 dataset](https://github.com/openforcefield/qca-dataset-submission/pull/259) is active (but making no progress)?

pavankum:
@jchodera this is the full optimization dataset and the other one is just single points. This was submitted for the purpose of comparing conformer selection strategies, MD based conformers versus regular toolkit generated conformers.

pavankum:
@dotsdl we can lower the priority on this one and after all single point datasets are evaluated we can push this

jchodera:
> @jchodera this is the full optimization dataset and the other one is just single points. This was submitted for the purpose of comparing conformer selection strategies, MD based conformers versus regular toolkit generated conformers.

Ah, thanks! I didn't catch that!

> @dotsdl we can lower the priority on this one and after all single point datasets are evaluated we can push this

That sounds reasonable!

PR #255, opened 2021-11-16
Folmsbee-Hutchison Conf. Benchmark

pavankum:
## New Submission Checklist

- [ ] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [ ] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [ ] Ready to submit!


pavankum:
@jthorton I concatenated all the geometries (mol files) from their repo (https://github.com/ghutchis/conformer-benchmark/tree/master/geometries) and tried to create an optimization dataset but OE failed with `"Programming error: OpenEye atom stereochemistry assumptions failed."`, and rdkit also fails with an isomorphic check. If you have some time tomorrow please take a look.
Another question is whether we want opt set or single points, they mentioned that they didn't optimize these geometries fully so I am leaning towards an opt set but then we can't compare directly to their results, any thoughts?

jthorton:
@pavankum Thanks for starting this one it looks like an issue with the connection information in the mol files, but loading from the [smiles file](https://github.com/ghutchis/conformer-benchmark/blob/master/SMILES/molecules.smi) seems to work fine. I would suggest creating the molecules from the smiles file and then for each optimised conformer do an isomorphism check while not comparing bond order then put the conformer in the correct order and add it to the molecule. The paper does mention that these were optimised using `B3LYP-D3BJ /def2-svp` so I think single points should be fine. I think it would be good to also try and preserve the names of the molecules if possible in the dataset as well. Would you like me to have a go at this?

pavankum:
> @pavankum Thanks for starting this one it looks like an issue with the connection information in the mol files, but loading from the [smiles file](https://github.com/ghutchis/conformer-benchmark/blob/master/SMILES/molecules.smi) seems to work fine. I would suggest creating the molecules from the smiles file and then for each optimised conformer do an isomorphism check while not comparing bond order then put the conformer in the correct order and add it to the molecule. The paper does mention that these were optimised using `B3LYP-D3BJ /def2-svp` so I think single points should be fine. I think it would be good to also try and preserve the names of the molecules if possible in the dataset as well. Would you like me to have a go at this?

@jthorton Sure, that'd be great. Thank you for looking into it.

PR #258, opened 2021-12-07
SPICE DES370k Monomers Revised v1.1 (wcombine = False)

pavankum:
Revised submission with addition of `'wcombine': False` to the keywords and bumping the version to v1.1 since the first submission went through compute.


dotsdl (approved):
This looks great and ready to go @pavankum! Merging!

PR #257, opened 2021-12-07
SPICE Solvated Amino Acids (wcombine = false)

pavankum:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [x] Ready to submit!

PR #259, opened 2021-12-07
SPICE Dipeptides revised v1.1 (wcombine = False)

pavankum:
## New Submission Checklist

- [x] Ready to submit!


jthorton:
@pavankum please resolve the conflict and pull in master to run validation with qcsubmit 0.3.0. Otherwise looks great!

dotsdl (approved):
Thank you @pavankum! Merging!

jchodera:
@pavankum Has something gone wrong here? The dataset lost a bunch of COMPLETE entries, and is now parked at 0 compete for many days:
image

pavankum:
@jchodera David [updated the error cycling scripts](https://github.com/openforcefield/qca-dataset-submission/commit/ce81defe211a400a7a246d2f9dbf29334ea2261d) recently to correct a bug and I guess that's the reason for the mismatch in numbers, @dotsdl am I right?

jchodera:
@pavankum : Looks like something must be wrong here. These are single-point calculations so should be very fast to run, but there are still 0 complete after 23 days of running?

pavankum:
@jchodera these calculations are still in queue, throughput seems to be around 10000 or more single points per day with the current compute resources for small molecules, the error cycling reports are generated automatically every day irresepctive of any actual calculations being done and that's why some of these may show zero runs complete. 
Please let us know if you want to prioritize this over other spice sets. Right now, monomers set (#258) is complete, half of pubchem_set1 (#260) is done, and solvated amino acids (#257) seem to consume lot of time and resources. And there are other openff datasets too Chapin's 2D torsion scans (#262), and expanded gen2 opt sets (#263). Another factor, `def2-tzvppd` being a larger basis the runs are a bit slower than our openff default specification.
cc: @dotsdl

jchodera:
@pavankum : Ah! I think I've just misunderstood how the queue works. Rather than assigning a weight to each project (like Folding@home), it must be more like a first-in-first-out queue. My apologies for the misunderstanding---you folks have the right idea about how to prioritize resources.

jchodera:
Wow, that's a lot of errors!

@pavankum: Any idea what went wrong? There were so many errors the traceback was truncated, so it's impossible to tell if this is some sort of issue with the calculation or the configuration of workers.

pavankum:
@jchodera Yeah, I agree, will dig into it and report back.

dotsdl:
Re-tagging to `openff-defunct`; this dataset is probably contributing to our immediate storage issue on public QCArchive. CC: @bennybp

jchodera:
@dotsdl : You should feel free to erase it if you need to restart it without saving wavefunctions!

dotsdl:
Thanks @jchodera; deletion is not technically supported by the current QCFractal server, but @bennybp have discussed this as a DB-surgery solution as a last resort. We do have deletion as a supported operation in the `next` QCFractal server iteration upcoming.

PR #268, opened 2022-01-26
repo size reduction

### LINKED ISSUE #267: Repository Size
## Description

The size of the repository seems to be growing very large (on the order of GBs), which makes it very slow to clone on slower internet connections, and in general doesn't seem fully sustainable as larger and larger submissions are added. 

Is it worth considering moving some of the larger files to LFS rather than storing the directly in the repo? This would include large dataset and pdf files.

pavankum:
To resolve #267 added files greater than 15MB in size to LFS.


dotsdl:
I don't have experience with `git-gc`, but after getting the big files moved LFS I think we could try to use it. Probably best as as separate PR, or may need to be done forcefully since I think it counts as a kind of rebase (rewrites history)?

SimonBoothroyd:
Thanks @pavankum - I've not really used LFS before but on the surface at least this looks good to me!

pavankum:
I think migrating files to LFS is not retroactive and the history remains pretty large, I am checking [this thread on stackoverflow](https://stackoverflow.com/questions/47199828/how-to-convert-a-file-tracked-by-git-to-git-lfs) and there seems to be a package called [BFG_repo_cleaner](https://rtyley.github.io/bfg-repo-cleaner/) that might be useful for us. I am not familiar with handling such large repo changes so we can discuss this in the next qcf meeting. 
CC: @j-wags @mattwthompson to have their eyes on this sometime.

j-wags:
I also don't have a ton of experience with git LFS but I think it's probably a good solution. I'd like to better understand how we'd need to use it, and what the price is for our use cases. @dotsdl will look into it this sprint if he has spare bandwidth. https://trello.com/c/V9k7lO5I

dotsdl:
@pavankum can you share the sequence of commands you ran to create this PR? It's not clear to me from the diff or looking at the branch locally that this gives us the desired result. Want to make sure I understand it before merging.

dotsdl:
@pavankum is there a `.gitattributes` file in your repo from the operations you did? I think we'll need that committed and pushed here as well.

pavankum:
@dotsdl I did 
```
git lfs track filename
git add filename
```
for all the files > 15 MB. This migrated the files to LFS and I got pointers in place of those.

pavankum:
yeah, I have it, somehow I am having an issue locally with `Encountered 1 file(s) that should have been pointers, but weren't:`, trying to resolve that.

pavankum:
> It's not clear to me from the diff or looking at the branch locally that this gives us the desired result.

Yeah, I have the same doubt, I can see the files online on the repo tagged with `Stored with Git LFS`
image
and yet the local copies remain when I switch the branch, hmm...

PR #276, opened 2022-02-01
OpenFF ESP Industry Benchmark Set v1.0

SimonBoothroyd:
## New Submission Checklist

- [X] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [X] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!

- [ ] Ready to submit!


SimonBoothroyd:
@dotsdl will a submission of this size cause issues?

SimonBoothroyd:
Running on lilac I wasn't getting great throughput, so thought it would be best to punt to the full QCF instance and let whatever compute is available churn away at it.

pavankum:
@SimonBoothroyd: discussed this with @dotsdl this morning, currently optimization sets don't store wavefunctions, so do you want to run the full optimization and use the final geometries later to do single points with wavefunctions? If so, we can push this to compute now. 
The second part might be tricky, Ben's back of the envelope estimate for storing wavefunctions is around 300GB for single points of this set, so that may still be delayed depending on how fast the storage issue is resolved.

SimonBoothroyd:
@pavankum thanks for checking! Yeah my plan is to not store optimization wavefunctions, but rather take the final conformers, prune any where the connectivity changes / h-bonds form, then create a new single point set from these.

pavankum (approved):
Perfect! Thank you for the clarification. I don't think it is necessary to regenerate without the store_wavefunction keyword since it is not considered in opt sets. Will squash merge this.

pavankum:
Looks like this is too big for the submission action, we can wait for the point release of qcfractal, otherwise can chop this into two parts, or submission via local resources.

pavankum:
Just checked the dataset status, and it seems good with expected number of records.

pavankum:
@SimonBoothroyd I think the last few are consistent errors, shall we move this to end of life?

dotsdl:
Switched off hourly error cycling; this will now cycle once daily. Let me know if you want more frequent error cycling again and I can switch it back on.

dotsdl:
@SimonBoothroyd looks like errors may be systematic at this point. Do you have what you need from this dataset? Can we move it out of error cycling?

PR #278, opened 2022-02-11
Prod env update

dotsdl:
Using python 3.8, latest of all packages except:
- psi4, as next is major version 1.5, which we may want to validate
  before production deployment
- pytorch, as torchani 2.2 uses it, and still no new relese of that
  package

### DIFF
--- a/devtools/conda-envs/backlog.yaml
+++ b/devtools/conda-envs/backlog.yaml
@@ -4,7 +4,7 @@ channels:
   - conda-forge
 
 dependencies:
-  - python =3.7
+  - python =3.8
   - pip
   - pip:
       - PyGithub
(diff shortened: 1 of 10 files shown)


dotsdl:
@pavankum what is your stance on sticking with psi4 1.4.1 vs. using 1.5? I'm inclined to keep using 1.4.1 at this time if we see no compelling reason to advance to 1.5.

pavankum:
@dotsdl Yeah, I am good with v1.4.1, new updates that fix the issue with range separated hybrids will be in v1.5.1 I guess (the next version) so we're good for now with using `wcombine=False`.

pavankum (approved):
Thank you for the feature enhancements in packages, updated envs and the testing!! Merging as is!

pavankum:
for qcportal I think the version should be 0.15.8

dotsdl:
Ah thanks!

PR #279, opened 2022-02-13
update python version on workflows

pavankum:
github actions are failing because the workflow scripts had python version set to 3.7


chapincavender (approved):
LGTM. Thanks @pavankum!

PR #281, opened 2022-02-18
Bump dipeptide 2-D TorsionDrive version to v2.1

chapincavender:
Bumped version to v2.1 in dataset name.


pavankum (approved):
Thank you @chapincavender! I think this is good to go!

pavankum:
@chapincavender I think this one reached 25/26 now, is the remaining one the problematic torsion with ARG? Can we move this to end-of-life?

pavankum:
Moving to end of life, remaining calculation is a consistent error with geometric convergence

PR #286, opened 2022-03-08
Add OpenFF ESP Industry Benchmark Set v1.1

SimonBoothroyd:
## New Submission Checklist

- [x] Ready to submit!

- [ ] Ready to submit!


jthorton:
@SimonBoothroyd thanks, I have added the MBIS flag and our other defaults such as WBO as we may also need these for a gnn ? if not feel free to remove them apart from MBIS of course :)

submissions/2022-02-02-OpenFF-ESP-Industry-Benchmark-Set-v1.1-single-point/generate.py
```diff
@@ -0,0 +1,96 @@
...
+        base_molecule._conformers = [m.conformers[0] for _, m in records]
+
+        dataset.add_molecule(
+            index=base_molecule.to_smiles(
+                isomeric=True, explicit_hydrogens=False, mapped=False
+            ),
+            molecule=base_molecule,
+            attributes=MoleculeAttributes.from_openff_molecule(base_molecule),
+            extras=base_record.extras,
+            keywords=record.keywords,
```
jthorton, line 80:
```suggestion
            keywords=base_record.keywords,
```

SimonBoothroyd:
@jthorton this seems to be using the default QC spec instead of the custom HF one specified, can you double check if I'm missing something here?

jthorton:
@SimonBoothroyd I pulled down the dataset in the PR and it is showing one spec which is the OpenFF default so it looks like the spec was not updated correctly.

submissions/2022-02-02-OpenFF-ESP-Industry-Benchmark-Set-v1.1-single-point/generate.py
```diff
@@ -0,0 +1,97 @@
...
+        driver=DriverEnum.energy,
+        metadata=Metadata(
+            submitter="simonboothroyd",
+            long_description_url=(
+                "https://github.com/openforcefield/qca-dataset-submission/tree/master/"
+                "submissions/"
+                "2022-02-02-OpenFF-ESP-Industry-Benchmark-Set-v1.1-single-point"
+            )
+        ),
+        qc_specs={
```
jthorton, line 41:
```suggestion
        qc_specifications={
```

jthorton: @SimonBoothroyd it looks like the field name is incorrect.

SimonBoothroyd: Ah great catch!

dotsdl:
Hmmm...not clear to me why the `dataset.json.xz` file isn't going to LFS instead of getting committed and pushed here. Trying to think of ways to characterize this further before we proceed.

SimonBoothroyd:
Sounds good - let me know if there's anything to do to help get this merged as this a a slight block on some science I want to get started 🙂

PR #289, opened 2022-03-22
Screened initial conformers with bad geometry using qcelemental guess_connectivity

chapincavender:
Errors in ~~#281~~ #277 were caused by initial conformers with bad geometry. These conformers have been screened using `guess_connectivity` from qcelemental, similar to [this filter](https://github.com/openforcefield/openff-qcsubmit/blob/ceecd8ee78b0a9eff5702057fd0e963a0148fb87/openff/qcsubmit/results/filters.py#L707) from qcsubmit. This dataset was not prepared using qcsubmit because it contains TorsionDrives on the same molecule with different constraints on non-driven dihedrals, which would be skipped in the qcsubmit deduplication step. If this dataset had been prepared with qcsubmit, the ConnectivityFilter could have been applied as a workflow component with no additional code. Thus, I feel like a general filter to solve this problem (changes in connectivity during conformer generation) already exists in qcsubmit and could be used for datasets that don't need to avoid deduplication.

Checklist for this resubmission:
- [x] Bumped dataset version in dataset creation, dataset filename, and submission README
- [x] Added changelog to submission README
- [x] Updated metadata (e.g. mean number of conformers) in submission README
- [x] Ready to submit!


pavankum:
Thank you @chapincavender, this looks great! Before merging this one naive question, you're skipping those grid points where the initial geometry was bad, and you're okay with all scans not having the same number of grid points, right?

pavankum (approved):
Everything looks great! If you're okay with the previous comment please go ahead and squash merge this.

chapincavender:
> Thank you @chapincavender, this looks great! Before merging this one naive question, you're skipping those grid points where the initial geometry was bad, and you're okay with all scans not having the same number of grid points, right?

@pavankum I'm not skipping grid points, I'm just failing to provide a Sage-optimized initial conformer for some grid points. I think these grid points will still be computed, and the optimizations will be seeded by adjacent grid points. Is that correct?

pavankum:
hmm, afaik if we provide a single molecule, with grid spacing and dihedral indices to constrain, torsiondrive procedure can perform the scan, [here is one example with qcengine](https://github.com/MolSSI/QCEngine/pull/305#issue-936896576). I hope it works even if we skip a grid point, @dotsdl any comments?

dotsdl:
> hmm, afaik if we provide a single molecule, with grid spacing and dihedral indices to constrain, torsiondrive procedure can perform the scan, [here is one example with qcengine](https://github.com/MolSSI/QCEngine/pull/305#issue-936896576). I hope it works even if we skip a grid point, @dotsdl any comments?

Yes, `torsiondrive` should generate adjacent grid points from provided scans and fill in as it propagates from computed optimizations. I don't know for sure how well it will work for these long sidechains, but worth a shot to see.

chapincavender:
@dotsdl @pavankum thanks, I agree that this is the expected behavior. Should I squash and merge?

pavankum:
Thank you @dotsdl! @chapincavender Sure, please go ahead!

pavankum:
Moving this to end-of-life since #290 supersedes

PR #292, opened 2022-05-24
OpenFF multiplicity correction torsion drive data v1.0

jmaat:
These are the torsion drive and optimization training datasets for multiplicity corrections in open force field. There is an issue with generating the torsion drive dataset, which I think is due to "symmetry". Only 1 of the 99 molecules gets to the final dataset. For the optimization dataset, 26 of the 99 molecules are filtered. I am not sure why this is the case. @pavankum 

## New Submission Checklist

- [ ] Ready to submit!

- [ ] Ready to submit!


pavankum:
@jmaat it looks the dihedral indices are not matched to the atoms you intend to drive the torsion for, `dataset.filtered_molecules` shows this filter being applied: `'UnconnectedTorsionRemoval'` because of the index mismatch.
In the pickle file you are using in the first step, can you add mapped smiles instead of the regular smiles so that the torsion indices you have in the dict correspond to the correct atoms.

And for the optimization set failures, at first glance it looks like an issue with conformer generator which tried MMFF and UFF forcefields and failed to generate conformers for some molecules. Let me check if we can resolve it with xtb-crest or other way of generating conformers.
Apart from that, you can leave the function arguments empty here, `factory = OptimizationDatasetFactory()`,  so that our default qc specification is populated instead of the wb97 functional.

jmaat:
I tried using canonical isomeric mapped smiles with explicit hydrogens by generating the smiles from the offmols using this function: 
`offmol.to_smiles(isomeric=True, explicit_hydrogens=True, mapped=True)` 
Here's an example of the smiles strings I am generating: 
`[H:5][O:1][S:2](=[O:3])[O:4][H:6]`
I am still having the same issue with the torsion drive dataset using these smiles strings. @pavankum

jmaat:
I added two scripts I use to generate torsion_drive_data.pickle in the torsion drive directory.
-run_v3-smarts_imp.py : is just to run the script
-smarts_impropers_angle_update_ff_3.py : this has the script which generates the pickle file and converts the smiles string to offmol. I’m sorry these scripts are really messy. I made 3 comments with “pavan” in them in this script so you can see the important places.

pavankum:
@jmaat the errors were due to index mismatches, I cleaned up the dataset generation scripts and included everything in one file. Please review them and let me know if the tagged torsions and parameters are consistent or not. 

Also, separated out the opt set submission to another PR #293, since having different dataset types in one PR may pose a problem with errorcycling, please review that as well. 

Also, please check the details in README files and add any extra description as you deem fit.

jmaat:
@pavankum Thank you, I double checked the parameters and the highlighted indices, and it looks good to go. I approved the PR

pavankum (approved):

pavankum:
@jmaat I looked at the remaining 7 torsion scans which are showing consistent failures, here are the molecules and the parameters they're exercising in your new FF,
 
image
Torsion_parameter: t133. For this, I think we shouldn't scan all 24 grid points since it is impossible to rotate the in-ring torsion at all those angles. A reduced set of grid points would work or another molecule without an in-ring torsion if the smarts can match. 

image
Torsion_parameter: t122d. For this, some of the grid points with constrained torsion angles seem to be sterically congested. Here is a look at the starting geometry for one of the failing optimizations
image





I think that's the case with the remaining failures below as well,

image
Torsion_parameter: t122d

image
Torsion_parameter: t122d

image
Torsion_parameter: t122a

image
Torsion_parameter: t122a

@jmaat If you have alternatives to these molecules we can add them to the current dataset and bump the version, please let me know.

pavankum:
Here's a notebook I used to debug the failures
[multiplicity_correction_dataset_failed_mols.tar.gz](https://github.com/openforcefield/qca-dataset-submission/files/8857245/multiplicity_correction_dataset_failed_mols.tar.gz)

PR #300, opened 2022-07-08
Added rna-bgsu-dataset-v1.0

kntkb:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [x] Ready to submit!

## Compute Expansion Checklist

- [ ] Compute expansion filename matches pattern `compute*.json`; may feature a compression extension, such as `.bz2`
- [ ] QCSubmit validation passed
- [ ] Ready to submit!

PR #301, opened 2022-07-12
Pr/300

pavankum:


pavankum:
LGTM! Ported the updates from #300 to here.

pavankum:
@dotsdl @chapincavender I reviewed this one and it is good to go, please approve this when you have time. This is a port of #300.

chapincavender (approved):
LGTM! Everything looks good to go for computation. A few minor changes for the README, but these shouldn't block submission and can be addressed later. Just wanted to document them while I'm thinking about it.

* I don't see that the main README for the repo was changed to point to this dataset. Was that in a previous PR?
* In the manifest, for the pickle files, change the description from "the main input for the dataset" to something more descriptive that differentiates them.
* In the manifest for `dataset.json.bz`, change "gradient dataset" to "basic dataset"
* In the manifest for `dataset.smi`, change "peptide" to "nucleic acids" or similar.

pavankum:
Thank you very much @chapincavender! Updated the readme files now.

pavankum:
@kntkb Calculations with the `openff-default` QC specification finished, and remaining 114 are persistent SCF convergence errors. For the other QC spec (`wb97m-d3bj/def2-tzvppd`) the basis set is pretty large and the completion rate is much slower, so it may take a while to finish with current compute resources. 

Accessing the data is a bit tricky for single points datasets, so if you want to look at the data we already have I am attaching a jupyter notebook for your reference (please rename the file to `*.ipynb` after download). Please do note the split of `B3LYP-D3BJ` calculation into `B3LYP` and `D3BJ`, and how to access each of those outputs. The same reason why you would see 8978 calculations for `default` in the above bot report, which is two times 4489.
 
[rna_dataset_access.ipynb.txt](https://github.com/openforcefield/qca-dataset-submission/files/9192066/rna_dataset_access.ipynb.txt)

kntkb:
Hi @pavankum! Perhaps we can stop the recalculation for the incomplete datasets since the most important ones are already complete. I can look into the incomplete datasets and prepare a better initial structure if necessary.

pavankum:
Thank you for the feedback @kntkb. Sure, will move this to end-of-life then.

PR #315, opened 2022-08-18
valid_scf_props argument added under compute spec check

pavankum:
Validation was failing for compute spec addition since the argument extras was missing from `create_spec_report()` under `check_compute_request()`


dotsdl (approved):
Thanks for this @pavankum! Merging!

PR #320, opened 2022-10-29
Update queued-submit.yaml

pavankum:
importlib_metadata is resolving to some outdated version, pinning it here.

### DIFF
--- a/devtools/conda-envs/queued-submit.yaml
+++ b/devtools/conda-envs/queued-submit.yaml
@@ -13,4 +13,4 @@ dependencies:
   - PyGithub
   - pandas
   - tabulate
-  - importlib_metadata =4.11.4
+  - importlib_metadata =5.0.0
--- a/devtools/conda-envs/queued-submit.yaml
+++ b/devtools/conda-envs/queued-submit.yaml
@@ -13,3 +13,4 @@ dependencies:
   - PyGithub
   - pandas
   - tabulate
+  - importlib_metadata =4.11.4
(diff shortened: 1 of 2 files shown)


j-wags (approved):
(not blocking) Hmm, I don't think the toolkit pin should be necessary, since qcsubmit currently [pins](https://anaconda.org/conda-forge/openff-qcsubmit/files) to `openff-toolkit<0.11`. And I'm wondering if the `importlib_metadata` pin should be migrated up to OFFTK, since that's where the first error occurred. 

Either way, while I spend my time wondering, we should probably get the validation action working again, so this PR should be good to merge. Let's just remember to unpin it later. Could you add comments justifying the pins so we know when we can remove them?

pavankum:
Thanks for the review @j-wags. Yeah, the pin on toolkit is not needed, updated that one. Also, pinned `importlib_metadata` to the  newest version since the trouble was when conda resolved it to `v1.5` which was incompatible with `importlib-metadata`, the package with hyphen. I think this should work fine, it worked locally for me.

pavankum:
>And I'm wondering if the importlib_metadata pin should be migrated up to OFFTK, since that's where the first error occurred.

I think there was a version of it which had `entry_points()` as a dict, and they were maintaining backward compatibility until `v5.0.0` which deprecated it completely, so a pin of `>= 3.6` may help. [This comment](https://github.com/python/importlib_metadata/issues/409#issuecomment-1266160825) is helpful.

pavankum:
Merging this for now.

PR #323, opened 2022-11-18
Upping qcengine to 0.25.0, qcelemental to 0.25.1 for non-psi4 prod envs

dotsdl:
We cannot use 0.25.0 alongside psi4 until a new release of psi4 is made.
psi4 1.6.1 only allows for qcengine >=0.23.0 <0.25.

However, we want to proceed with qcengine 0.25.0 for OpenMM use, since
this is compatible with openff-toolkit >= 0.11.0.

### DIFF
--- a/devtools/prod-envs/qcarchive-user-submit.yaml
+++ b/devtools/prod-envs/qcarchive-user-submit.yaml
@@ -13,4 +13,4 @@ dependencies:
   - openff-qcsubmit
   - openff-toolkit 
   - openeye-toolkits
-  - qcelemental =0.25.0
+  - qcelemental =0.25.1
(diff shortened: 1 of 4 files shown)


dotsdl:
@pavankum this will introduce a difference in QCEngine/QCElemental versions we are running between the `psi4` env and other envs. Is this tolerable to you until we have a new release of `psi4`?

pavankum:
@dotsdl Thanks for the env updates David! Sure, this setup works. I think we should pin the `importlib_metadata` package to `> 4.0.0` since that's been causing some trouble lately.

pavankum (approved):

pavankum:
Looks like we may not need that pin, I don't see the package with underscore anymore
```
  + importlib-metadata                    5.0.0  pyha770c72_1            conda-forge/noarch         23 KB
  + importlib_resources                  5.10.0  pyhd8ed1ab_0            conda-forge/noarch         29 KB

```

PR #324, opened 2022-11-21
Attempt to perform Docker build with mamba to avoid memory issues from conda

dotsdl:
See title; hitting what appear to be memory issues in build for openff docker image.
Switching to mamba to address this.


dotsdl:
Looks like this works as expected: https://github.com/openforcefield/qca-dataset-submission/actions/runs/3518648089

Should be good to merge @pavankum if satisfied.

pavankum (approved):
Thanks a lot @dotsdl!

PR #333, opened 2023-11-02
Multi-Br ESP dataset

lilyminium:
New submission testing the new QCSubmit version.

## New Submission Checklist

- [x] Ready to submit!


dotsdl:
Awesome, thanks for making this @lilyminium! I'll get started on #331 to fix the validation CI first so we can get its reporting.

dotsdl:
@lilyminium getting the following when fetching this branch:

```
Switched to a new branch 'multi-br-v1.0'
Encountered 1 file that should have been a pointer, but wasn't:
	submissions/2023-11-02-OpenFF-multi-Br-ESP-Fragment-Conformers-v1.0/dataset.json.bz2
```

Can you check that you have `git-lfs` installed? It should be the case that on commit and push a smudge file gets created in the place of the actual file. It won't matter too much for this case (`dataset.json.bz2` is only 184kB), but does help to keep the size of the repo down over time.

lilyminium:
Yes sorry, you're right, I did this on lilac where I did not have git-lfs installed. Ah... I re-committed and pushed on my laptop with lfs, which I *think* was supposed to fix it, but I'm not familiar enough with lfs to check 😅

dotsdl:
@lilyminium all good! It now shows up as an LFS file in the diff:
image

dotsdl:
@lilyminium looks like we have validation working again! Can you check to see what it's giving makes sense to you? I'm checking myself as well.

lilyminium:
Thank you @dotsdl! All the validation properties look good on my end that I can tell.

dotsdl (approved):
Looks great @lilyminium! Merging!

dotsdl:
Working on fixing the above submission failures today.

dotsdl:
@jthorton raised that we likely don't want wavefunctions for each optimization of this set. I'm working on the submission infrastructure still, so we could revise this before it makes its way to QCArchive.

I've asked @bennybp if we can now get wavefunctions for only the final structure for the optimization; in the past this wasn't possible, but not sure if this has changed.

PR #335, opened 2023-11-10
(Rebake of) Multi-Br ESP dataset

dotsdl:
Rebake of #333, setting trajectory protocol for each optimization to only save 'final' results, including wavefunctions.

PR #336, opened 2023-11-14
Update lifecycle automation to be compatible with PortalClient

dotsdl:
This PR updates the lifecycle automation to be compatible with QCFractal > v0.50.

We were able to simplify substantially, removing many of the workarounds we previously relied on. New workardounds may be necessary as we identify the need.


dotsdl:
Going to self-merge so we can get going with this and debug any issues using #335. Created PR here to easily reference the changes.

PR #337, opened 2023-11-16
Updating prod envs for compute deployments

dotsdl:
Attempting to use latest packages, pinning where appropriate to create some degree of determinism over time for the most important packages when building prod instances from these environments.

### DIFF
--- a/.github/workflows/prod-envs-docker.yml
+++ b/.github/workflows/prod-envs-docker.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         env-name:
-          - qcarchive-worker-openff-ani
+          #- qcarchive-worker-openff-ani
           - qcarchive-worker-openff-openmm
           - qcarchive-worker-openff-psi4
           - qcarchive-worker-openff-xtb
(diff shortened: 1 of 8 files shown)


dotsdl:
@bennybp can I get a spot-check on these envs for suitability as compute manager environments? We'll also use these for supporting our Docker-based manager deployments.

bennybp:
I think those are reasonable, at least at first glance. The only thing might be that you don't need explicit `gau2grid` anymore with the conda-forge psi4 packages.

dotsdl:
Awesome, thank you @bennybp!

dotsdl:
Got the envs building successfully now: https://github.com/openforcefield/qca-dataset-submission/actions/runs/6938585368

`pytorch` isn't happy with python 3.10 or 3.11, and the mamba solver errors are baffling. Choosing 3.11 and we get complaints that 3.10 is possible, but choosing 3.10 gives complaints that 3.11 is possible. Since it's unclear if there's any demand for `torchani` in the near or distant future, I've disabled build of this env in both prod env deployment and Docker build.

dotsdl:
Will revisit `torchani` env if desired in the future.

PR #339, opened 2023-11-22
Set pytorch to 2.1.0 for torchani env

dotsdl:
@lilyminium noted that using pytorch 2.1.0 appears to solve with conda. Trying it here.


dotsdl:
Looks like this works! https://github.com/openforcefield/qca-dataset-submission/actions/runs/6952576998/job/18916360001

Merging!

PR #344, opened 2024-01-31
Update and re-run dataset preparation example

mattwthompson:
I used this dataset (`cp ../../submissions/2021-06-22-OpenFF-BCC-Refit-Study-COH-v2.0/molecules.smi .`) locally.

- [x] Drop OEChem, which is not used in this notebook and can cause conflicts with QCPortal
- [x] Re-run notebook


lilyminium (approved):
LGTM -- thanks for the update @mattwthompson!

PR #345, opened 2024-01-31
Add supplemental torsion coverage dataset

ntBre:
## New Submission Checklist

- [ ] Ready to submit!


lilyminium (changes requested):
Thanks for opening this, @ntBre! Unfortunately the molecules here include several elements we don't cover: As, B, Si. Just skimming the info, some of them also look like they could be difficult -- e.g. quite large molecules up to 833 MW, and charges down to -6. Do all these chemistries need coverage, or would it be possible to fragment the molecules to isolate the torsion of interest a bit more?

ntBre:
Yes, I was surprised to see As today, but I did see that B and Si were present in some of the other datasets in this repo and thought those might be okay. In any case, I'm happy to remove all three elements if you want. I was also concerned about the charges. Should I filter out anything below -2? And what is an acceptable molecular weight? I was previously filtering by number of atoms instead. They all have fewer than 80 atoms, so perhaps the maximum is being raised by an arsenic atom.

ntBre:
Simply filtering out molecules with these elements and charges only removes 12 molecules. I'm going to re-run the notebook and see what comes out now.

lilyminium:
@ntBre you could filter by charge/MW, sure -- another approach would be fragmenting molecules on rotatable bonds or using something like rdkit's `Recap.RecapDecompose` (I have a script for this), or even doing it manually ( e.g. in the screenshot molecules below, there's quite a lot of tail around the driven torsion that could be snipped off, and the right hand side probably doesn't need all those sulfate groups).

Screenshot 2024-02-01 at 9 13 07 am

lilyminium:
This also caught my eye -- is this parameter meant for aromatic rings, or is it a case of a single bond SMILES accidentally matching the kekulisation/aromaticity perception somewhere between OpenFF/OpenEye/RDKit?

Screenshot 2024-02-01 at 9 17 34 am

lilyminium:
I'm also getting a warning when I fetch your branch (the same one as [here](https://github.com/openforcefield/qca-dataset-submission/pull/333#issuecomment-1797884644)) -- could you please check that you have `git-lfs` installed?

ntBre:
I installed git-lfs as the readme suggested yesterday and cloned after exporting `GIT_LFS_SKIP_SMUDGE=1`. Do I have to set that for all of my git operations?

ntBre:
> These also caught my eye -- are these parameters meant for aromatic rings, or is it a case of a single bond SMILES accidentally matching the kekulisation/aromaticity perception somewhere between OpenFF/OpenEye/RDKit? (I'll stop updating with more, but a check for aromatic rings might be worthwhile here!)

I can't really answer the parameter part of this question since the images aren't annotated with the parameters they are supposed to match. What would be the check for aromatic rings to use?

lilyminium:
You could try looking for [ring aromaticity directly](https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html) with `[aR:1]`, or (a bit more tiresome) check all r5 rings. I'm guessing you could get back the parameter ids by searching the `all.smiles` file for aromatic rings, or visually checking the smirks of the torsion parameters you're covering here.

lilyminium:
Hm, I can't quite remember how I set it up... I think when I accidentally pushed it as well, I just recommitted with lfs. It *should* read the `.gitattributes` at the top level of the repo, it shouldn't need an individual `.gitattributes` file :/

lilyminium (commented):
Glad to see this being updated! I get the feeling that many of these torsions will be very stiff, but it'll be good to get data on them. Just a couple minor notes but otherwise LGTM!

* Just checking, are the molecules here postprocessed through RecapDecomposition, or selected via clustering? If so, could you please add some notes on that?
* The date of this directory is now somewhat out-of-date, could you please change it to a February date?
* There's just a conflict (in the main README, I think)

I wonder why the validation bot isn't running -- it might be because of the draft status?

ntBre:
Thanks again for the feedback! I think I have addressed all of your points, and I also fixed the rebase mistake I made yesterday. Hopefully it is ready for a new review!

lilyminium (approved):
Fantastic, the description of the process is very comprehensive and the links to descriptions of the algorithms are great, thank you! LGTM!

PR #346, opened 2024-02-13
Update README.md

amcisaac:
Added starting workers to readme checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The lifecycle process is described below, with [bracketed] items indicating the 
         - if successful, will move card to state ["Error Cycling"](https://github.com/openforcefield/qca-dataset-submission/projects/1#column-9577365); add comment to PR
         - if failed, will keep card queued; add comment to PR; attempt again next execution
      
-    - [Human] submit worker jobs on a server to begin compute
+    - [Human] Submit worker jobs on a server to begin compute. If using Nautilus, carefully monitor utilization and scale down resources as jobs finish.
 
 4. COMPLETE, INCOMPLETE, ERROR numbers reported for `Optimizations`, `TorsionDrives`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,7 +114,9 @@ The lifecycle process is described below, with [bracketed] items indicating the
     - [GHA]  [`lifecycle-backlog`](.github/workflows/lifecycle-backlog.yml) will move PR card to state ["Queued for Submission"](https://github.com/openforcefield/qca-dataset-submission/projects/1#column-9577335) if merged and in state ["Backlog"](https://github.com/openforcefield/qca-dataset-submission/projects/1#column-9577334)
     - [GHA]  [`lifecycle-submission`](.github/workflows/lifecycle-submission.yml) will attempt to submit the dataset
         - if successful, will move card to state ["Error Cycling"](https://github.com/openforcefield/qca-dataset-submission/projects/1#column-9577365); add comment to PR
-        - if failed, will keep card queued; add comment to PR; attempt again next execution 
+        - if failed, will keep card queued; add comment to PR; attempt again next execution
+     
+    - [Human] submit worker jobs on a server to begin compute
```
lilyminium, line 119: Thanks for adding this! I wanted to suggest adding more specific notes, e.g. about monitoring and scaling up/down the workers as necessary, but changed my mind as it doesn't quite fit here as part of the dataset "life-cycle". That  being said, could you please add a note to also stop the workers when there the dataset has finished computing? I think that still fits in and a task something that could easily slip someone's mind.

lilyminium (changes requested):

README.md
amcisaac: Sure--let me know what you think! I added a vague note about monitoring Nautilus usage, but agree that specific details might not fit here

PR #356, opened 2024-06-14
Add another torsion multiplicity torsion drive supplement

ntBre:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [ ] QCSubmit validation passed
- [ ] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!

PR #361, opened 2024-06-17
Fix number of conformers and add main readme entry to #356

ntBre:
This fixes the lingering issues with #356.


lilyminium:
Thanks @ntbre! I don't think the dataset and notebook need updating in the PR, as there shouldn't be any major changes to them (well, especially the dataset, I suppose the notebook was updated). Could you please just remove the dataset updates just to avoid potential issues with triggering new jobs? I'm not sure that it would but better safe than sorry.

ntBre:
Oops, I thought I reverted the changes to the dataset, but I must have opted not to. It should be fixed now!

lilyminium (approved):

lilyminium:
Thanks @ntBre, LGTM!!

PR #364, opened 2024-07-17
OpenFF Phosphate Torsion Drives v1.0

ntBre:
This is a torsion drive dataset generated from the list of SMILES I got from Patrick via Lily. It's currently quite large, as I finally figured out that qcsubmit has a `get_symmetry_group` function for providing the `symmetry_group` argument to `TorsionIndexer.add_torsion`. Thus, for a first pass, I added a torsion drive for each torsion matching the lipid parameters Patrick sent (t2, t3, t4, t45, t46, t50, t95, t123a, t159, and t160). Combining this with `workflow_components.StandardConformerGenerator(max_conformers=10)` leads to 5121 conformers. 3 of the provided SMILES also fail to match any of these parameters, so they are currently not included in the dataset. There are also some high negative charges (-5, -4) and a max molecular weight of 785.55, so we may need to do some additional filtering before submission.

## New Submission Checklist

- [ ] Ready to submit!

## Todos
- [x] Update dataset README after finishing changes
- [ ] Update main README


lilyminium (approved):
LGTM -- thanks @ntBre for shepherding this one through!

j-wags:
I'm... not sure why this dataset is so cursed at the moment. Each job takes a huge amount of resources (workers are 16 cores x 32 GB) and still take 1+ hours per job, and even then they're mostly returning errors. I'm concerned that error cycling is now outpacing these failing jobs, so the ones that might complete aren't being reached at all.

So I'm temporarily removing the `tracking` label so these stop getting error cycled, and we can let the waiting queue empty out and see if any of these are successful in a day or two. Either way, I think this dataset is nearing its final state, and unfortunately about 25% of TDs won't be properly finished (but judging be the number of opts, most of them will have at least one structure per scan point)

PR #368, opened 2024-08-14
Update automation environments

jthorton:
Update the automation environments to use the latest qcsubmit versions which should allow the MLPepper dataset to submit, see failed attempt [here](https://github.com/openforcefield/qca-dataset-submission/actions/runs/10380312200/job/28739926732).


lilyminium (approved):
LGTM. I checked QCSubmit releases and there only seems to be one obscure API break so I'll merge and keep an eye on actions to see if anything breaks :-)

jthorton:
Thanks @lilyminium that worked!

PR #369, opened 2024-08-27
Deprecate DockerHub and Anaconda.org use

### LINKED ISSUE #328: Switch Docker image push to use GHCR, deprecate DockerHub
Since GHCR is now a fairly mature option for Docker image hosting (we've been using this for [alchemiscale](https://github.com/openforcefield/alchemiscale), we should try switching over to this from DockerHub. This will simplify management of images, keeping them alongside the repo itself, and avoid potential fees from DockerHub (though with our open source status, we have so far avoided these).

j-wags:
- [x] Fixes #328 
- [x] Removes use of anaconda.org
- [x] Updates docs


j-wags:
I was able to successfully deploy ddx-containing workers for the MLPepper dataset using the stuff from this PR, so I think this is basically done modulo docs updates. But I'm going to focus on fixing #370 first since the reporting in that will give us some insight into whether the newly deployed workers are successful at scale.

Issue #370, opened 2024-08-30
Update automation for new GH project boards


j-wags:
Something about lifecycle management actions is currently failing with the following error. I'm able to reproduce locally. 

```
Run python ./management/lifecycle.py --states "Error Cycling" --set-priority --set-computetag
  python ./management/lifecycle.py --states "Error Cycling" --set-priority --set-computetag
  shell: /usr/bin/bash -l {0}
  env:
    MAMBA_ROOT_PREFIX: /home/runner/micromamba
    MAMBA_EXE: /home/runner/micromamba-bin/micromamba
    CONDARC: /home/runner/work/_temp/setup-micromamba/.condarc
    GH_TOKEN: ***
    QCA_USER: ***
    QCA_KEY: ***
Traceback (most recent call last):
Found 11 with the 'tracking' label
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 987, in <module>
    main()
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 931, in main
    board = _get_full_board(repo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 813, in _get_full_board
    proj = [proj for proj in repo.get_projects() if proj.name == "Dataset Tracking"][0]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
Error: Process completed with exit code 1.
```

j-wags:
It appears to have something to do with how the old project boards are repo-specific (ex `https://github.com/openforcefield/qca-dataset-submission/projects/2`) and the new ones are accessed on the org level (ex `https://github.com/orgs/openforcefield/projects/2`).

j-wags:
Ah, and pygithub doesn't support the new style of projects yet https://github.com/PyGithub/PyGithub/issues/2140

j-wags:
To get REST access to the project board contents using the V2 API, I had to get the project ID. It wasn't at all clear how to do this in the browser, so I made a token at the org level with read access to org:project, then ran

```python
import subprocess

cmd = [
    "curl", "--request", "POST",
    "--url", "https://api.github.com/graphql",
[...]

PR #371, opened 2024-09-04
Fix error cycling

j-wags:
Update project board fetching to use GraphQL interface so we can run error cycling. 

This doesn't handle board modification yet.


j-wags:
This partially resolves #370 by getting error cycling to at least run on MLPepper (though it times out in the process, it does somehow reset the tasks and post statistics to the PR). 

I haven't tested the other lifecycle steps and assume there are more gremlins to be found. This also doesn't re-implement moving cards between columns. But error cycling is by far the most important thing (and I'm tired of seeing it failing). So I'm going to merge this to get things to a less-broken state.

PR #372, opened 2024-09-07
Nagl2 esp timing benchmark

amcisaac:
## New Submission Checklist

- [x] Ready to submit!

- [ ] Ready to submit!


lilyminium (commented):
Generally LGTM -- thanks @amcisaac! Could you please add an entry to the overall README?

amcisaac:
On my end it looks like I updated it, but maybe I added it to the wrong place? I put it at the end of `Basic Datasets`
Screen Shot 2024-09-16 at 5 28 05 PM

lilyminium:
Ah sorry my bad -- I completely missed it. LGTM!

lilyminium (approved):

lilyminium:
Oh dear. FWIW the `--- Too many errors; truncated here ---` is a QCA message -- querying the QCPortal should pull down actual output for more helpful diagnosis.

amcisaac:
Looks like Psi4 can't do VV10 dipole polarizabilities, causing all the calculations to crash, based on the error `Fatal Error: Vx: RKS cannot compute VV10 Vx contribution.`

I just copied the list of properties from the MLPepper dataset since I figured it couldn't hurt--Do we need/want the dipole polarizabilities? Josh mentioned they included that because they are looking into a polarizable charge model, which to my knowledge we aren't working on. Happy to either 1) re-submit without requesting that or 2) change functional if we do

j-wags:
Ah, the MLPepper dataset ran with psi4 1.9 and needed an additional dependency (pyddx). So these were probably crashing because they were landing on our `openff`-label workers, which (I assume) for provenance reasons have psi4 1.8 and no pyddx. 

We did the MLPepper dataset on the `isambard` tag so I could ensure that only workers with newer deps accept those tasks. 

So we'll want to retag this set and try sending it to newer workers made using the `psi4-ddx` image.

I'll give this a shot right now but am just heading offline for the week, so apologies if I leave it in a messy state.

amcisaac:
This doesn't have ddx, so I think the reason is due to the VV10 gradients not being implemented (as mentioned [here](https://forum.psicode.org/t/sapt-dft-wb97x-v-error/1342)), which means we can't calculate the dipole polarizabilities. We're going to re-submit with PBE0

amcisaac:
That's good to know though, as the next step is to submit a version with ddx for solvent!

j-wags:
Ah, nice. Thanks for the clarification, I'll keep an eye open for your resubmission.

PR #375, opened 2024-09-10
Patch lifecycle to avoid interacting with project boards overly

lilyminium:
This PR patches lifecycle.py to give up on moving things between board columns and just submit the datasets. It's meant to be a temporary, easier to review fix than #377.


lilyminium:
Apologies for the messy dangerbot actions -- I refactored lifecycle.py to use GraphQL, but then split that out into a separate PR #376. This is now just a bandaid patch to try to get datasets submitted before #376 can get merged. We will have to manually move the card to the Error cycling to get that to trigger, I believe.

lilyminium:
The passing CI run was triggered manually here: https://github.com/openforcefield/qca-dataset-submission/actions/runs/10807416097/job/29978159402

lilyminium:
I think this is now ready for review.

j-wags (approved):
Looks great, thanks for updating this and apologies that this held up submissions!

lilyminium:
Thanks for the review @j-wags!

PR #378, opened 2024-09-11
OpenFF Sulfur Optimization Training Coverage Supplement v1.0

ntBre:
This is a dataset for increasing coverage on the groups outlined [here](https://openforcefield.atlassian.net/wiki/spaces/FF/pages/2606465079/Increasing+dataset+coverage#To-do). As mentioned in the README, these were selected from ChEMBL 34 based on the SMIRKS patterns in `sulfur.dat`. If this looks good, I have an additional 5 molecules for each label that I will submit as a separate dataset for use in benchmarking with the same input scripts.

## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!

PR #384, opened 2024-09-18
 OpenFF Sulfur Hessian Training Coverage Supplement v1.0

ntBre:
This is a single-point dataset for computing Hessians at the final structures from #378.

## New Submission Checklist

- [ ] Ready to submit!


submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/generate.py
```diff
@@ -0,0 +1,111 @@
...
+opt = OptimizationResultCollection.from_server(
+    client,
+    datasets=["OpenFF Sulfur Optimization Training Coverage Supplement v1.0"],
+)
+
+print(f"Retrieved {opt.n_results} results")
+
+with portal_client_manager(lambda _: client):
+    recs = opt.filter(
+        RecordStatusFilter(status=RecordStatusEnum.complete)
```
lilyminium, line 30: Could you also please add a ConnectivityFilter for records where the connectivity may have changed?

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/README.md
```diff
@@ -0,0 +1,57 @@
+# OpenFF Sulfur Hessian Training Coverage Supplement v1.0
```
lilyminium, line 1: Thanks so much @ntBre  for putting this together!

I'm going back and forth on the naming a bit. In the past Hessian datasets afaict were all named the same as the OptimizationDatasets. The main advantage of continuing to do this are that it's clear which OptimizationDatasets have had Hessians calculated; the obvious downside is that it's a bit confusing to have a singlepoint Hessian dataset called "OpenFF Sulfur Optimization Training Coverage Supplement v1.0".

Renaming it as you've done here clears up that confusion, and the name is similar enough to the optimization dataset that they definitely look linked. This is also what we do for ESP calculations. I double checked how `OptimizationDataset.to_basic_result_collection(driver='hessian')` works, since that's how we pull down Hessians in the MSM calculation script. It [looks like](https://github.com/openforcefield/openff-qcsubmit/blob/90b325296f8eed07199ea7d0e395a3aacd84958a/openff/qcsubmit/results/results.py#L606-L607) this doesn't rely on the Dataset name, so that should all be fine. 

The only slight downside is that when we curate data for a FF fit, we only specify the optimization datasets to download and it'll be less clear which of them have Hessians. Typing this out makes me lean towards it not being a big deal, but just checking if you think it would negatively impact your fitting workflow when you change or expand training datasets, @ntBre and @amcisaac? Would be great to get your thoughts on this and I'll defer to what you decide.

ntBre: Ah thanks for bringing this up, I didn't realize they had the same names in the past. I thought it would be convenient to have Hessian in the name for searching through the main table of datasets, but I see how it would be important to know where the structures came from! I did include that in the README and description at least, if we decide not to rename.

As far as fitting goes, as you said, I just use `to_basic_result_collection` on a single `OptimizationResultCollection`, which is already stitched together from multiple named datasets, so it's not really clear to me which ones have Hessians at that point anyway.

I don't have strong feelings either way but would lean toward leaving it just because it's more convenient not to have to rename. I'm happy to take care of it if you and Lexie prefer it renamed, though.

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/generate.py
ntBre: Will do!

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/README.md
amcisaac: I'm a bit confused--are the existing "Optimization" datasets with hessians the results of a geometry optimization that saved hessians at each step, or it's a single point datasets with Hessians calculated that is just called "Optimization"? I would think a single point dataset that saves Hessians should be called "Hessian" like this or "Basic" to correspond to the other single points. It seems super confusing to have single point datasets mixed in with optimization datasets. However, if this is how we've done it in the past, I do see how changing the convention could also cause confusion. 

As far as the workflow, it sounds like the two of you have looked into it and decided that it wouldn't affect things. If we convert all the optimization datasets to basic result collections anyway, I don't think it should matter.

Personally I think it's better to change the convention going forward to something more easy to understand such as calling it a Hessian dataset, and make a note in the README or something. I think it's especially for new people as the org grows, I feel like we rely a lot on the knowledge of previous contributors for things like this, and I would have never checked in the optimization datasets if I was looking for single points. But I don't have a strong preference either way; if we keep it the same and call it an Optimization dataset, I'd also make a note of that somewhere in the README.

lilyminium: > are the existing "Optimization" datasets with hessians the results of a geometry optimization that saved hessians at each step, or it's a single point datasets with Hessians calculated that is just called "Optimization"

The latter -- if you look at the [Smirnoff Coverage Set 1 for example](https://github.com/openforcefield/qca-dataset-submission/blob/1f3842b0767d7c3462225f9eda115a45bfd62d76/submissions/2019-06-25-smirnoff99Frost-coverage/04_create_hessian_dataset.py#L31) you can see where the name is the same. Just as far as I can tell hessians have been computed in the same directory as their opt sets so maybe it was easier.

> I feel like we rely a lot on the knowledge of previous contributors for things like this, and I would have never checked in the optimization datasets if I was looking for single points.

Right now the Hessian datasets are listed in BasicDataset table, so they are discoverable, and up until now they have been named after the Opt dataset they descended from. That being said I tried to download the "SMIRNOFF Coverage Set 1" basic set just now using `qcsubmit` and got a KeyError from QCSubmit[^1], so maybe older datasets don't include all the nice things we expect and would be best interacted with through the Opt dataset anyway. (Downloading directly through the PortalClient itself works fine.)

Anyway, sounds like we're generally in agreement to name hessian datasets differently from Optimization ones, and maybe be clearer about documentation / where they descend from. Thanks both for your input!


[^1]: KeyError trace:
	```
	KeyError                                  Traceback (most recent call last)
	Cell In[6], line 1
	----> 1 collection = BasicResultCollection.from_server(
	      2     client=qc_client,
	      3     datasets=[
	      4        "SMIRNOFF Coverage Set 1"
	      5     ],
	      6     spec_name="spec_1",
	      7 )
	
	File ~/micromamba/envs/qcsubmit-050/lib/python3.11/site-packages/openff/qcsubmit/results/results.py:366, in BasicResultCollection.from_server(cls, client, datasets, spec_name)
	    363     datasets = [datasets]
	    365 # noinspection PyTypeChecker
	--> 366 return cls.from_datasets(
	    367     [
	    368         client.get_dataset("singlepoint", dataset_name)
	    369         for dataset_name in datasets
	    370     ],
	    371     spec_name,
	    372 )
	
	File ~/micromamba/envs/qcsubmit-050/lib/python3.11/site-packages/openff/qcsubmit/results/results.py:325, in BasicResultCollection.from_datasets(cls, datasets, spec_name)
	    321     cmiles = molecule.extras.get(
	    322         "canonical_isomeric_explicit_hydrogen_mapped_smiles"
	    323     )
	    324 if not cmiles:
	--> 325     cmiles = entry.attributes[
	    326         "canonical_isomeric_explicit_hydrogen_mapped_smiles"
	    327     ]
	    328 if not cmiles:
	    329     print(f"MISSING CMILES! entry = {entry_name}")
	
	KeyError: 'canonical_isomeric_explicit_hydrogen_mapped_smiles'
	```

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/generate.py
lilyminium: Thank you!!

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/README.md
lilyminium: I took the liberty of slightly modifying the table entry to explicitly mention the optimization dataset by name while fixing the merge conflict -- hope that's ok with you Brent.

j-wags:
This was [failing](https://github.com/openforcefield/qca-dataset-submission/actions/runs/10981489341/job/30488577490) to be processed in our machinery (log excerpt below). Same with #383. I think there's a problem with the dataset names not matching somewhere - Maybe try fixing and move back to error cycling?

```
...
Processing PR #385
Setting computetag to 'pyddx'
Processing dataset 'submissions/2024-09-18-OpenFF-NAGL2-ESP-Timing-Benchmark-v1.1/dataset.json.bz2'
Processing PR #384
Setting computetag to 'openff'
Processing PR #383
Setting computetag to 'openff'
Processing dataset 'submissions/2024-09-13-OpenFF-Sulfur-Optimization-Benchmarking-Coverage-Supplement-v1.0/dataset.json.bz2'
    main()
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 1000, in main
    submission.execute_state(board=board,
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 171, in execute_state
    return self.execute_errorcycle(pr_card, pr_state,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 265, in execute_errorcycle
    results.append(ds.execute_errorcycle(reset_errors=reset_errors,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/lifecycle.py", line 487, in execute_errorcycle
    ds = client.get_dataset(dataset_type, dataset_name)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/error-cycle/lib/python3.11/site-packages/qcportal/client.py", line 233, in get_dataset
    ds = self.make_request("post", f"api/v1/datasets/query", Dict[str, Any], body=body)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/error-cycle/lib/python3.11/site-packages/qcportal/client_base.py", line 416, in make_request
    r = self._request(
        ^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/error-cycle/lib/python3.11/site-packages/qcportal/client_base.py", line 381, in _request
    raise PortalRequestError(f"Request failed: {details['msg']}", r.status_code, details)
qcportal.client_base.PortalRequestError: Request failed: Could not find optimization dataset with name 'OpenFF Sulfur Optimization Benchmarking Coverage Supplement v1.0' (HTTP status 400)
```

submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0/README.md
amcisaac: Oops, sorry, I read this but never responded. I see, if it's listed in the Basic Dataset section that makes a lot of sense too. I think both ways make sense though I still think calling it a Hessian set is a bit more straightforward

PR #395, opened 2024-10-22
Update STANDARDS.md

amcisaac:
Updating the STANDARDS document to be in line with our current practices as well as outline things we'd like to ensure are included moving forward.

### DIFF
--- a/STANDARDS.md
+++ b/STANDARDS.md
@@ -14,41 +14,37 @@ We distinguish between standards for the datasets (i.e. the actual data), and th
 - Coordinates
 - Total charge
 
-## Each dataset must have the following information:
+## Each dataset's metadata/entries must have the following: 
 - Name
-- Version (NEW)
+- Version 
 - A short description
 - A long description
 - A link/URL/reference pointing to the provenance to reproduce dataset (e.g. the GH submission folder)
-- A changelog (NEW)
+- A github submitter username
+- Each entry has canonical isomeric explicit hydrogen mapped SMILES
+
+## Each dataset subdirectory in QCA Dataset submission must have the following information:
+- Full environment used to generate dataset
+- Dataset generation script including the following:
+        - How conformers were generated
+        - Origin of SMILES strings
+	- Enumerated stereo flag (True/False)
+	- Enumerated tautomers flag (True/False)
+	- Enumeration keywords/settings
+
+ - A descriptive README file, as described below
+
+## Each dataset README must contain the following information:
+
+- All information specified in the dataset
+- A changelog (if there are changes)
 	- The changelog is a python-like dictionary of the form `{ version: entry }`
 	- Each entry in the changelog has the python dictionary form `{ person: str, date: str, description: str }`
-- A github submitter username
-- The name of the person who selected/sourced the molecules (NEW)
-- A description of the meaning of the entry/molecule keys/names (NEW)
-- Each entry has canonical isomeric explicit hydrogen mapped SMILES
-- Provenance info of CMILES generation
+- The name of the person who selected/sourced the molecules 
+- A description of the meaning of the entry/molecule keys/names
 - A set of elements that the dataset contains
-- A set of charges that the dataset contains (NEW)
-- The mean and max molecular weight of molecules the dataset contains (NEW)
-- Enumerated stereo flag (True/False) (NEW)
-- Enumerated tautomers flag (True/False) (NEW)
-- Enumeration provenance info (NEW)
-- Computation blacklist (known failures) (NEW)
-- Dataset status flags (NEW):
-	- Complete: `COMPLETE` `INCOMPLETE`
-	- State: `DONE` `WORKING` `PAUSED`
-	- Compliance: `V3` `NONE`
-	- COMPLETE/DONE/V3; all molecules were successful
-	- INCOMPLETE/DONE/V3; some molecules were not calculated successfully, and won't be retried
-	- INCOMPLETE/WORKING/V3; in progress
-	- INCOMPLETE/DONE/NONE; the dataset does not conform to the standards, and can't be fixed
-	- INCOMPLETE/WORKING/NONE; the dataset does not conform to the standards, but is working anyways
-	- INCOMPLETE/PAUSED/NONE; the dataset does not conform to the standards, and calculations have been suspended
-
-## Each dataset README must contain the following information
-
-All information specified in the dataset
+- A set of charges that the dataset contains
+- The mean and max molecular weight of molecules the dataset contains 
 
 ## Each revision must use the following procedure:
 
--- a/STANDARDS.md
+++ b/STANDARDS.md
@@ -58,7 +58,7 @@ As a rule, we never modify existing datasets, so any changes must be made using 
 
 A revision means creating a new dataset based on an existing one, with the intent of fixing/improving it.
 
-- The changelog must be copied from the current version (if it exists), and a new entry added for the new version. The changelog should describe any changes between the new version and previous version, including QCAIDs of any records that were removed or modified, with an explanation as to why they were removed.
+- The changelog must be copied from the current version (if it exists), and a new entry added for the new version. The changelog should describe any changes between the new version and previous version, including QCArchive record ids of any records that were removed or modified, with an explanation as to why they were removed.
 - The dataset version information is updated, and the description updated if necessary
 - The README is updated
 - A notebook or script used to manipulate the dataset and generate the revision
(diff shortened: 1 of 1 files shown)


STANDARDS.md
```diff
@@ -2,65 +2,69 @@
 This file outlines the standards and data requirements needed for submitting a dataset to QCArchive.
 This ensures that we have a consistent data model for downstream processes.
 
-STANDARDS version: 3 (adopted 2020.12.11)
+STANDARDS version: 4 (adopted 2024.11.XX)
```
amcisaac, line 5: Question: do we want to update the version number? Nothing has ever conformed to version 3 as far as I can tell from the submissions directory, so we could continue using it, but I thought it might be better to bump it so there is no confusion.

STANDARDS.md
```diff
@@ -2,65 +2,69 @@
...
+- A set of charges that the dataset contains
+- The min, mean and max molecular weight of molecules the dataset contains 
 
 ## Each revision must use the following procedure:
 
 A revision means creating a new dataset based on an existing one, with the intent of fixing/improving it.
 
-- The changelog must be copied from the current version, and a new entry added in the new version
-- The dataset version information is updated
+- The changelog must be copied from the current version (if it exists), and a new entry added for the new version. The changelog should describe any changes between the new version and previous version, including QCAIDs of any records that were removed or modified, with an explanation as to why they were removed.
```
amcisaac, line 61: Are people happy with this description of a change log? The dictionary format sounded hard to parse and unnecessarily complicated. But I'm happy to put it back if people feel strongly.

STANDARDS.md
```diff
@@ -2,65 +2,69 @@
...
+- The dataset version information is updated, and the description updated if necessary
 - The README is updated
-- A notebook or record of the lines of python used to manipulate the dataset responsible for the revision
-	- Each file of record should be named with the version that it first addresses, e.g. `submit-v3.0`
-		- If `submit-v3.0` also includes a `v3.1 and v3.2` update, then a new file for `v3.2.1` would be `submit-v3.2.1.py`
-	- Python notebooks should have version changes in order, and can be run incrementally.
+- A notebook or script used to manipulate the dataset and generate the revision
```
amcisaac, line 64: Are revisions supposed to be submitted in their own directory, or in the directory of the original dataset? This makes it sound like the revision would be submitted in the directory of the original dataset. I don't think we're currently doing this, so we should be on the same page about it moving forward, and update the standards to indicate it explicitly.

STANDARDS.md
```diff
@@ -70,7 +74,7 @@ A revision means creating a new dataset based on an existing one, with the inten
 
 ## Dataset naming and versioning
 
-Each dataset shall be versioned.
+Each dataset shall be versioned according to `vX.Y.Z`. X refers to the major version, Y to the minor version, and Z to a micro version.
 - The naming of a dataset should have the following structure:
 
     `"OpenFF <descriptive and uniquely-identifying name> v<version number>"`
```
amcisaac, line 80: See question above about whether we want to update versioning to start at 4, or keep it at 3

STANDARDS.md
```diff
@@ -79,17 +83,16 @@ Each dataset shall be versioned.
 
 * The major version shall indicate the STANDARDS that the dataset conforms to. Datasets which are not intended to conform to any STANDARDS should start with 0, e.g. `"v0.1"`. 
 
-* Datasets with versions starting with `"v1.x"` and `"v2.x"` do not follow any official STANDARDS, and thus should be considered `"v0.x"`.
+* Datasets with versions starting with `"v1.x"` and `"v2.x"` do not follow any official STANDARDS, and thus should be considered `"v0.x"`. 
 
 - A minor version change (e.g. `"v3.1"`) represents a minor addition and/or fixes problems:
-	- Adding molecules
+	- Adding molecules or removing/replacing records with an underlying problem
 	- Adding compute specifications
```
amcisaac, line 90: Do we really want compute specifications to require a version bump? You already have to specifically request the compute spec you want, so I don't think there's any confusion. 

Additionally, our existing procedure for compute expansions doesn't allow modification of the dataset name, so we would have to modify or eliminate it in favor of submitting a completely new dataset for each compute spec update.

STANDARDS.md
```diff
@@ -79,17 +83,16 @@ Each dataset shall be versioned.
...
 	- Errors/bugs in the molecule specification
 	- Changes necessary to adhere to the STANDARDS (i.e. changes necessary to placate the NONE compliance status)
 
 - A micro version change (e.g. `"v3.1.1"`) represents a cosmetic change, or a change that is based on dynamic information that does not change the underlying data:
```
amcisaac, line 94: Given that we've eliminated the compute blacklist in favor of filtering out problems and bumping the version, do we want to use a micro version? I can't think of a cosmetic change that would require us to specify a version, for example updating the status or description wouldn't need a new version IMO

STANDARDS.md
```diff
@@ -116,13 +119,13 @@ Pre-submission filtering:
 Post-submission filtering:
```
amcisaac, line 119: Given the standards above for versions, doing any post-submission filtering would require a minor version bump. Is that how we intend this to be used? That would mean that any released dataset would be v3.1, and nothing would ever be v3.0 unless all the stereochemistry, topology, etc was perfect and didn't need to be filtered.

STANDARDS.md
lilyminium: SGTM!

STANDARDS.md
lilyminium: Based on its location in this doc, it sounded like the dict format for a CHANGELOG was intended to fit Pythonically in the metadata of the dataset somewhere. As a markdown file this makes sense to me.

lilyminium, line 61:
```suggestion
- The changelog must be copied from the current version (if it exists), and a new entry added for the new version. The changelog should describe any changes between the new version and previous version, including QCArchive record ids of any records that were removed or modified, with an explanation as to why they were removed.
```

Just suggesting to explicitly refer to QCA record IDs instead of by shorthand :-)

lilyminium (approved):
LGTM!! I've left comments on the discussion points you raised.

amcisaac:
OK great, thanks! I'll make these changes, and wait to merge until next week in case anyone else wants to comment

STANDARDS.md
amcisaac: Trevor mentioned that it's referring to the [`UnperceivableStereoFilter`](https://docs.openforcefield.org/projects/qcsubmit/en/stable/_modules/openff/qcsubmit/results/filters.html#UnperceivableStereoFilter), which writes the molecule to an SDF file and tries to read it back in. I think this has to be applied after an optimization, since the QM structure and the structure from the conformer generation may differ. But for a single point dataset, it could be done before.

I updated the text to mention that, and specify which other filters can be used.

STANDARDS.md
```diff
@@ -115,14 +111,17 @@ Pre-submission filtering:
 
 Post-submission filtering:
 
-- OpenFF toolkit ingestion with strict stereo checking in RDKit
-- Hydrogen bonding
+- Stereochemistry is preserved (e.g. QCSubmit's `UnperceivableStereoFilter`)
+- CMILES (topology) change (e.g. QCSubmit's `ConnectivityFilter`)
+- Calculation ran successfully (e.g. QCSubmit's `RecordStatusFilter`)
+- Hydrogen bonding (for torsion drives only) (e.g. QCSubmit's `HydrogenBondFilter`)
```
ntBre, line 117: It might be nice to link to the code or, likely preferably, the docs for these filters.

STANDARDS.md
```diff
@@ -70,30 +72,24 @@ A revision means creating a new dataset based on an existing one, with the inten
...
 
-- The first submission of a dataset will have a version `"v3.0"`
+  Where the words in the dataset should be separated by spaces.
+
+- The first submission of a dataset will have a version `"v4.0"`
 
 * The major version shall indicate the STANDARDS that the dataset conforms to. Datasets which are not intended to conform to any STANDARDS should start with 0, e.g. `"v0.1"`. 
 
-* Datasets with versions starting with `"v1.x"` and `"v2.x"` do not follow any official STANDARDS, and thus should be considered `"v0.x"`.
+* Datasets with versions starting with `"v1.x"` and `"v2.x"` do not follow any official STANDARDS, and thus should be considered `"v0.x"`.
```
ntBre, line 86: This is the tiniest complaint imaginable, but there's a trailing space here.

ntBre (approved):
Looks great to me, thanks for your work on this! I made two very small comments about the text, but these are obviously non-blocking and don't affect the actual standards, which look great.

STANDARDS.md
lilyminium: Thanks @amcisaac!

PR #394, opened 2024-10-09
Add LIPID MAPS fragment optimization dataset

ntBre:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!

PR #408, opened 2024-11-19
Include molecule size suggestions in README

amcisaac:
Adding suggestion to split dataset into molecules less than 300 Da and more than 300 Da to the repo README

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,14 @@ Datasets must be submitted as pull requests.
    
 
 
-4. Create and activate a new conda env with basic submission-preparation requirements with:
+3. Create and activate a new conda env with basic submission-preparation requirements with:
     ```
     conda env create -f qca-dataset-submission/devtools/prod-envs/qcarchive-user-submit.yaml
     conda activate qcarchive-user-submit
     ```
 
 
-5. Choose a starting notebook and README based on the type of dataset you wish to submit:
+4. Choose a starting notebook and README based on the type of dataset you wish to submit:
     - [`OptimizationDataset`](./examples/OptimizationDataset/)
 
    Copy the notebook and README for the dataset you want into the directory you created.
@@ -51,7 +51,7 @@ Datasets must be submitted as pull requests.
    ```
 
 
-6. Start up a Jupyter notebook with your new notebook:
+5. Start up a Jupyter notebook with your new notebook:
     ```
     jupyter notebook qca-dataset-submission/submissions/YYYY-MM-DD-OpenFF-<DESCRIPTIVE-DATASET-NAME>-v1.0/generate-dataset.ipynb
     ```
@@ -60,11 +60,11 @@ Datasets must be submitted as pull requests.
    and make any other modifications as needed for your specific needs.
 
 
-7. Copy generated metadata components into README.
+6. Copy generated metadata components into README.
    Write a reasonably-detailed high-level description of the submission at the top.
 
 
-8. Commit the following files in the submission directory you made:
+7. Commit the following files in the submission directory you made:
     - your input files; please compress them if possible with e.g. `bzip2`
     - `generate-dataset.ipynb`
     - `dataset.pdf`
@@ -72,19 +72,19 @@ Datasets must be submitted as pull requests.
     - `dataset.json.bz2`
 
 
-9. Push your branch to Github:
+8. Push your branch to Github:
     ```
     git push origin <dataset-branch>
     ```
 
 
-10. Make a new PR for the branch.
+9. Make a new PR for the branch.
    Validation will run automatically on your `dataset.json.*` file, indicating any potential issues prior to submission.
    Ask for help if you see validation failures you do not understand.
    Ping a reviewer in the comments.
 
 
-11. Once reviewed and approved, your submission will be merged and submitted to QCArchive!
+10. Once reviewed and approved, your submission will be merged and submitted to QCArchive!
     Computations specified by the submission will be performed on OpenFF-managed compute resources.
 
 ### Creating a compute expansion
--- a/README.md
+++ b/README.md
@@ -29,15 +29,19 @@ Datasets must be submitted as pull requests.
    ```
    You will add all submission artifacts to this directory.
 
+   **If you have a wide distribution of molecule sizes in your dataset, please submit the dataset in multiple parts sorted by molecule size.**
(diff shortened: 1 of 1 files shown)


ntBre (commented):
I think this is reasonable given our recent discussions. However, I messaged Ben today about possibly tagging subsets of a single dataset with different compute tags. I think that would be a nicer solution, if you don't mind waiting on this until I hear back from him. I think splitting them as you described here is still likely to be the way to go, though.

amcisaac:
Sounds good, I think that would be more elegant if it's possible.

ntBre:
Ben said it's totally doable, so I'll try to figure out how to do it and maybe even experiment on the lipidmaps dataset tomorrow.

ntBre:
This turned out to be quite straightforward, and the basic mechanism is the same as the existing [code](https://github.com/openforcefield/qca-dataset-submission/blob/0e8e3e9d217af5e86c13994047168598829aab21/management/lifecycle.py#L608) for updating the compute tag on a dataset. Here's the example code I used to split up the LIPID MAPS dataset from #394. I'm happy to generalize this to include it somewhere in the repo if we want.

``` python
import os

if True:
    from qcportal import PortalClient

from openff.toolkit import Molecule
from openff.units import unit
from tqdm import tqdm


def get_tags(dataset, force_refetch=False):
    return {
        rec.task.tag
        for _, _, rec in tqdm(
            dataset.iterate_records(force_refetch=force_refetch),
            desc="Getting tags",
        )
        if rec.status != "complete"
    }


username = os.environ["QCA_USERNAME"]
password = os.environ["QCA_PASSWORD"]

client = PortalClient(
    "https://api.qcarchive.molssi.org:443",
    username=username,
    password=password,
)
ds = client.get_dataset(
    "optimization", "OpenFF Lipid Optimization Training Supplement v1.0"
)
ds.fetch_entries()

small = list()
medium = list()
large = list()
for entry_name, _, rec in tqdm(ds.iterate_records()):
    if rec.status == "complete":
        continue
    entry = ds.get_entry(entry_name)
    cmiles = entry.attributes[
        "canonical_isomeric_explicit_hydrogen_mapped_smiles"
    ]
    mol = Molecule.from_mapped_smiles(cmiles, allow_undefined_stereo=True)
    mass = sum(atom.mass for atom in mol.atoms)
    if mass > 600.0 * unit.dalton:
        large.append(rec.id)
    elif mass > 300.0 * unit.dalton:
        medium.append(rec.id)
    else:
        small.append(rec.id)

# print these to see the UpdateMetadata to check for errors
print(client.modify_records(small, new_tag="pyddx-small"))
print(client.modify_records(medium, new_tag="pyddx-medium"))
print(client.modify_records(large, new_tag="pyddx-large"))

# check that new tags are applied
print(get_tags(ds, force_refetch=True))
```

I can confirm that this changed all the tags, and I have workers picking up tasks from the `pyddx-small` tag right now. My only lingering question is whether this will be overwritten by any of the actions in the repo the next time the retagging action is run. Based on the fact that the Retag action is set to run 5 times per day, I think this will be the case, unfortunately.

@j-wags and @lilyminium could we remove the [cron](https://github.com/openforcefield/qca-dataset-submission/blob/0e8e3e9d217af5e86c13994047168598829aab21/.github/workflows/lifecycle-set-priority-computetag.yml#L9) entry from the Reprioritize/Retag action to enable this workflow? Or are there other consequences of that that I'm not considering? I've only ever run this action manually immediately after changing a tag, but I'm still new to this worker management.

On second glance, even the error cycling action passes the `--set-computetag` flag to the lifecycle script, so that would have to be removed too...

lilyminium:
Neat workflow @ntBre, that's pretty cool and nice work splitting up the lipid maps dataset! It'd be great to get this to a standard tool that we could use to split up datasets. Just a couple questions:

- is it possible to do this *before* dataset submission, instead of waiting until after a PR is merged and submitted to QCA? If so, could we build this into QCSubmit somehow?
- if not, could there be ways to fiddle with our existing tagging CI -- e.g. if a PR is tagged `compute-openff`, the action automatically appends it with `compute-openff-0-300`, `compute-openff-300-600`, .. -- I think we can set managers to pick up tasks that start with `openff-*` which would still pick up all of them if we need, or can set managers more specifically too

ntBre:
@lilyminium Thank you! 

I think the answer to your first question is no. Ben said you could [submit](https://github.com/MolSSI/QCFractal/blob/99e7b825342992e5facf5739e60e83244c799c80/qcportal/qcportal/dataset_models.py#L283) under different tags by specifying subsets of entry names, but I don't see any tag fields on [OptimizationDatasetEntry](https://github.com/MolSSI/QCFractal/blob/99e7b825342992e5facf5739e60e83244c799c80/qcportal/qcportal/optimization/dataset_models.py#L26), for example. I saw today that even records with the `complete` status don't have the `task` field, where the `tag` lives, so dataset entries before submission probably don't have this either. The `dataset.json` files from this repo that I've looked at only seem to have dataset-wide tags too. So I think it would require some kind of additional metadata to be read by the submission action in `lifecycle.py`.

Either way, I think your second suggestion is great. Building on that, instead of `compute-openff` becoming `compute-openff-0-300` and so on, it might be cool to have the user apply the tags `openff-0-300`... and have the workflow handle the splitting and tagging in that way. Tags without the trailing ranges could just be interpreted the normal way.

That would make the retag action more expensive since it would include retrieving the records and checking their masses like in my snippet above, but it would also neatly abstract all of this and keep it in the lifecycle script. And both the grouping and tagging could be controlled just by GitHub labels instead of partially at submission time.

I definitely think something like that would be nicer than having to share the username and password for anyone wanting to run a script like mine above!

If you like this idea, I can probably get a prototype working tomorrow.

lilyminium:
Hmm yeah I think something like what you're suggesting would be really cool and storing the required username/pws as org secrets a bit easier to handle than passing them around. Maybe the compute tag could be something like compute-xxx_mw-300-600 just to safeguard a bit further against compute tags with numbers in them for some reason? So yeah please do look into this! Although, as it's more of a "nice-to-have" for now it's still lower in priority than our current projects like the protein fits, so no need to have anything up quickly tomorrow if that'll disrupt what you currently have going on :-)

ntBre:
I think these changes are outdated in light of #412, but I can add something similar to the README about using the new tagging mechanism.

PR #409, opened 2024-11-20
NAGL2 training optimizations part 1

amcisaac:
## New Submission Checklist

- [x] Ready to submit!


amcisaac:
Unfortunately I haven't been able to find a way to suppress the Stereochemistry warnings in the `generate-dataset-part1.ipynb` notebook, making the output a bit unwieldly. I tried setting all possible loggers to `ERROR` and `CRITICAL`, neither of which suppressed them. I'm assuming it has to do with the multiprocessing or tqdm?

.gitattributes
ntBre, line 1: How large (and what) is this file? Without knowing anything about it, my immediate feeling is that it's either
1. Large and should just be bzipped, in which case it would match an existing pattern in .gitattributes
2. Small and doesn't need to be zipped or stored in LFS

Is the file both large and unable to be zipped?

submissions/2024-11-19-OpenFF-NAGL2-Training-Optimization-Dataset-Part-1-v4.0/README.md
```diff
@@ -0,0 +1,57 @@
+# OpenFF NAGL2 Training Optimization Dataset Part 1 v4.0
+
+## Description
+A dataset containing molecules from the [`MLPepper RECAP Optimized Fragments v1.0`](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2024-07-26-MLPepper-RECAP-Optimized-Fragments-v1.0)
+and [`MLPepper RECAP Optimized Fragments v1.0 Add Iodines`](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2024-10-11-MLPepper-RECAP-Optimized-Fragments-Add-Iodines-v1.0) datasets,
+with additional conformers and optimized at the OpenFF default level of theory (B3LYP-D3BJ/DZVP).
```
ntBre, line 6: This might be a minor point, but would you call these "additional conformers" or did you replace all of the conformers? From my reading of the notebook you only kept the CMILES and then regenerated conformers from scratch.

submissions/2024-11-19-OpenFF-NAGL2-Training-Optimization-Dataset-Part-1-v4.0/generate-dataset-part1.ipynb
ntBre, line 1: In box 18, where the molecules are added to the dataset, is there any reason to do it like this rather than passing the molecules to the dataset factory `create_dataset` method? Actually, on second glance, I think the `dataset1_factory` is not used after provenance is added to it, and all of the following operations use `dataset1` directly. Ah, now I see the factory appearing again in `add_molecule`. If there's no specific reason for this usage, I'd say call `create_dataset` and provide it the `molecules` keyword arg.

ntBre (commented):
I'm a bit biased, but I still think the subset tagging from #408 would be nicer than this manual splitting if we can resolve some of the issues with the automation. If you don't want to wait for that, this looks plausible to me otherwise, with the minor exceptions listed below.

.gitattributes
ntBre: Uh, this should have been a comment on the actual line changed, not the file itself. I know what .gitattributes is  and was asking about the addition of `mlpepper.json` :sweat_smile:

amcisaac: Oh, thanks, that's a good call. It's large but zippable

submissions/2024-11-19-OpenFF-NAGL2-Training-Optimization-Dataset-Part-1-v4.0/generate-dataset-part1.ipynb
amcisaac: Sure, I can do it this way, I was just copying the notebook from the dataset I was originally basing it on.

amcisaac:
I agree in the abstract but I'd like to get this submitted sooner rather than later, so I'm going to stick to this format for now

lilyminium (approved):
Thanks for putting this together @amcisaac and @ntBre for the thorough review! Largely LGTM although I agree with Brent's comments on mlpepper.json and wordings.

Up to you @amcisaac if you want to implement the individual compute tag methodology -- if you have no strong feelings I'd lean toward not doing so this time in the interests of getting things through sooner, plus it's not *unhelpful* to be able to download just the smaller/larger molecules. 
Approving to not block merges.

amcisaac:
Thanks for the feedback! I made the requested changes. I would just like for it to get going so I don't want to implement the compute tag methodology, though I definitely agree that it seems like an easier way to go.

PR #411, opened 2024-11-20
Don't retag automatically in lifecycle workflows

ntBre:
As covered in [this comment](https://github.com/openforcefield/qca-dataset-submission/pull/408#issuecomment-2489455218), retagging individual records is really handy for subdividing datasets based on molecule size to make requesting NRP workers easier, but the current lifecycle actions automatically retag ~6 times per day (5 from the retag action itself and 1 from error cycling), disrupting any manual tagging done as described in the comment.

I think these changes should prevent all automatic retagging. One remaining issue is that anyone manually running the retag action to update their dataset tag will still overwrite any of these subset tags, but at least that's likely to happen fewer than 6 times daily.

### DIFF
--- a/.github/workflows/lifecycle-error-cycle.yml
+++ b/.github/workflows/lifecycle-error-cycle.yml
@@ -52,4 +52,4 @@ jobs:
           QCA_USER: ${{ secrets.QCA_USER }}
           QCA_KEY: ${{ secrets.QCA_KEY }}
         run: |
-            python ./management/lifecycle.py --states "Error Cycling" --reset-errors --set-priority --set-computetag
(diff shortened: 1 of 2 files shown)


lilyminium (approved):
(more of a reply to the comment/s in the linked issue and Slack, but here seems like the best place to look back on) -- I'm not sure why the re-tagging action was set up to run 5x/day, but agree with @j-wags that it seems ok to just trial this and see if anything breaks.

In the future we can set up the re-tag action to just trigger on actual tagging/manual dispatch, if it isn't already.

ntBre:
Thanks Lily! And just in time before the next cron run at 00:00 UTC!

PR #412, opened 2024-11-22
Split dataset tags based on PR labels

ntBre:
This implements the feature discussed in #408, where labels beginning with `compute-` (this is enforced by the original code) and ending with `_mw-300-600-900`, just for example, the numbers can be anything, will be split up and tag separate portions of the dataset based on molecular weight. For example, `compute-openff_mw-100-200-300` will cause the creation of four tags: `compute-openff-100` for MW < 100, `compute-openff-200` for 100 <= MW < 200, `compute-openff-300` for 200 <= MW < 300, and `compute-openff-large` for anything larger than 300 Da. I wrote the splitting code in a generic way, so it should also handle  different numbers of groups other than 3, for example `_mw-300` or `_mw-100-200-300-400-500`.

For tags that don't match the pattern, the usual behavior should be preserved, which is good because the splitting implementation I added has to retrieve the dataset and iterate over its records to compute molecular weights, making it more expensive than the plain `dataset.modify_records` call.

I added a file `test_lifecycle.py` to test my implementation. I don't think this is run as part of normal CI, but it's at least helpful for local development.

### DIFF
--- a/management/lifecycle.py
+++ b/management/lifecycle.py
@@ -13,6 +13,7 @@ from collections import defaultdict, Counter
 from datetime import datetime
 
 from github import Github
+from qcelemental.models import Molecule
 
 QCFRACTAL_URL = "https://api.qcarchive.molssi.org:443/"
 
@@ -48,34 +49,28 @@ def parse_tags(compute_tag) -> tuple[list[float], str]:
     return list(reversed(ret)), tag.removesuffix("_mw") if len(ret) > 0 else tag
 
 
-def try_get_smiles(entry) -> str | None:
-    """Try to extract a SMILES from multiple fields in ``entry``. Based on code
-    from qcsubmit, see ``BasicResultCollection.from_datasets``, for example.
-    """
-    # this should cover either single-points or optimizations
+def try_get_molecule(entry) -> Molecule | None:
+    """Try to extract a qcelemental Molecule from multiple fields in ``entry``."""
     molecule = (
+        # this should work for singlepoints
         getattr(entry, "molecule", None)
+        # this for optimizations
         or getattr(entry, "initial_molecule", None)
     )
-    # this is how qcsubmit gets a molecule for torsion drives
-    if molecule is None:
-        return entry.attributes.get(
-            "canonical_isomeric_explicit_hydrogen_mapped_smiles"
-        )
-    # and this covers all the possibilities for finding a smiles in
-    # single-points or optimizations
-    return (
-        molecule.identifiers.canonical_isomeric_explicit_hydrogen_mapped_smiles
-        or molecule.extras.get(
-            "canonical_isomeric_explicit_hydrogen_mapped_smiles"
-        )
-        or entry.attributes.get(
-            "canonical_isomeric_explicit_hydrogen_mapped_smiles"
-        )
-    )
+
+    if molecule:
+        return molecule
+
+    # and this for torsiondrives
+    if (mols := getattr(entry, "initial_molecules")) is None or len(mols) < 1:
+        return None
+
+    return mols[0]
 
 
-def partition_records(ds, bins, include_complete=False) -> dict[int, list[int]]:
+def partition_records(
+    ds, bins, include_complete=False
+) -> dict[int, list[int]]:
     """Split up the records in ``ds`` based on the molecular weights (in Da) in
     ``bins``.
 
@@ -84,9 +79,6 @@ def partition_records(ds, bins, include_complete=False) -> dict[int, list[int]]:
     qcportal since complete records don't have tags), but for tests it's nice
     to be able to call this on a finished dataset
     """
-    from openff.toolkit import Molecule
-    from openff.units import unit
-
     # mapping of bin index to a sequence of record_ids
     ret: dict[int, list[int]] = defaultdict(list)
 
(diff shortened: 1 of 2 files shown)


management/lifecycle.py
```diff
@@ -26,6 +27,122 @@
...
+
+    return ret
+
+
+def set_mw_compute_tags(client, ds, compute_tag, include_complete=False):
+    bins, base_tag = parse_tags(compute_tag)
+    if len(bins) == 0:
+        print(f"Failed to parse molecular weight compute tags from {compute_tag}")
+        # TODO should this fall back on the normal setter then? we'd need to
+        # pass more arguments to make that possible
```
ntBre, line 126: Possibly I should add this to the function docstring, but this function is only supposed to be called when `SPLIT_TAG` matches (see `update_compute_tags`), so this `len(bins)` check *should* be overly cautious anyway. However, I thought it was still good to leave some kind of check in case of future changes.

lilyminium (changes requested):
Thanks for getting to this so fast @ntBre! I really like how it works. Since this function looks more expensive than the previous, I've added some suggestions for making it faster, and for adding docs so future users understand what's going on!

management/lifecycle.py
```diff
@@ -26,6 +27,122 @@
...
+
+    for entry_name, _, rec in ds.iterate_records():
+        if rec.status == "complete" and not include_complete:
+            continue
+        entry = ds.get_entry(entry_name)
+        if (cmiles := try_get_smiles(entry)) is None:
+            print(f"failed to get cmiles from {entry_name}")
+            continue
+        mol = Molecule.from_mapped_smiles(cmiles, allow_undefined_stereo=True)
+        mass = sum(atom.mass for atom in mol.atoms)
```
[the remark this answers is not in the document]
ntBre: Ah, thanks for the suggestion! I kind of assumed this was the case but never went to look. Somewhat surprisingly this slightly changed the numbers of records in each bin (I guess we're calculating the mass slightly differently?), but it's much nicer anyway.

management/lifecycle.py
```diff
@@ -26,6 +27,122 @@
...
+        client.modify_records(record_ids, new_tag=new_tag)
+
+
+def update_compute_tags(client, dataset, specification_names, new_tag, include_complete=False):
+    if SPLIT_TAG.search(new_tag) is None:
+        dataset.modify_records(
+            specification_names=specification_names,
+            new_tag=new_tag,
+        )
+    else:
```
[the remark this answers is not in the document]
ntBre: I think I handled both of these in the new `update_compute_tags` docstring!

management/lifecycle.py
```diff
@@ -26,6 +27,122 @@
...
+        # loop for the largest records (above any threshold)
+        found = False
+        for i, threshold in enumerate(bins):
+            if mass < threshold * unit.dalton:
+                ret[i].append(rec.id)
+                found = True
+                break
+
+        if not found:
+            ret[i+1].append(rec.id)
```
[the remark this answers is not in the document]
ntBre: Thanks!

ntBre:
Thanks for the review and the suggestions! Most of the "expense" I was worried about was actually waiting for QCArchive to respond (fetching records and entries instead of just calling `BaseDataset.modify_records`), but skipping the OpenFF `Molecule` construction and using numpy for the binning looks neater anyway.

Glad to hear the tests pass on your machine too! Most of my test runs take ~3.5 minutes, probably because I'm a lot closer to QCArchive.

management/test_lifecycle.py
```diff
@@ -0,0 +1,104 @@
...
+# avoid requiring PyGitHub to run tests
+sys.modules["github"] = mock.MagicMock()
+
+
+@pytest.mark.parametrize(
+    "input_tag,want",
+    [
+        ("compute-openff_mw-300-600", ([300.0, 600.0], "compute-openff")),
+        (
+            "compute-pr393_mw-300-600-900",
```
j-wags, line 15: (not blocking) To make this more readable, maybe replace the `pr393` with something obviously fake.

management/test_lifecycle.py
```diff
@@ -0,0 +1,104 @@
...
+
+    got = parse_tags(input_tag)
+    assert got == want
+
+
+@pytest.mark.parametrize(
+    "dsname,dstype,bins,want",
+    [
+        (
+            "OpenFF Sulfur Hessian Training Coverage Supplement v1.0",
```
j-wags, line 32: (not blocking) If these tests are taking a few minutes, maybe find a smaller dataset to ping?

management/test_lifecycle.py
```diff
@@ -0,0 +1,104 @@
...
+            "singlepoint",
+            [100.0, 200.0, 300.0],
+            [6, 220, 234, 439],
```
j-wags, line 35: (not blocking) It's kinda confusing to have these be "the number of things in each bin once you sort the bins by population". It would be more straightforward for it to be "the number of things in each bin, where the bins are sorted from lowest MW to largest".

j-wags (commented):
Thanks for the nice review @lilyminium - I added a few drive-by suggestions but I'll let you be primary reviewer (avoids ambiguity/so your approval is the final one)

management/test_lifecycle.py
ntBre: I'm going to skip this one for now since these aren't run in CI, but I agree in principle, especially if we add CI or end up having to modify this code again soon.

management/test_lifecycle.py
ntBre: Agreed. I updated the code to assert on the mapping of `bin_id` to `len(record_ids)` instead of just these sorted lengths.

lilyminium (approved):
LGTM -- thanks for the changes @ntBre!

ntBre:
Is this good to merge? I didn't want to rush into it if it broke anything while I was offline last week. Thanks again for the reviews!

PR #413, opened 2024-12-03
OpenFF Organometallics Exploratory Optimization Dataset

ntBre:
As the name indicates, this is an exploratory dataset to see how the openff-toolkit and qcsubmit handle organometallic molecules from the CCD, as well as to see if there are any issues on the QCArchive/Psi4 side. I was initially shooting for 100 molecules, but errors in the actual dataset generation phase filtered out 55 molecules (and one got filtered out for uniqueness, I think), leaving only 44. I think that will still be enough to see how well the calculations run and how easily we can retrieve the results.

## New Submission Checklist

- [ ] Ready to submit!


lilyminium (approved):
LGTM, thanks @ntBre. Just as notes, we had a meeting (with Jeff) and discussed:

* we won't dive into the `generate.txt` errors as we aren't intending to use the QCSubmit pathway for real data
* conformers will be ideally pulled from the CIF files from the real dataset instead of generated by RDKit
* this is a preliminary Psi4 test

PR #414, opened 2024-12-04
Only import qcelemental for type checking

lilyminium:
This fixes an issue where importing qcelemental at the top level breaks lifecycle.py in environments without qcel installed.


lilyminium:
Just going to quickly self-merge to get CI back running.

PR #416, opened 2024-12-09
Combine and filter NAGL2 Optimization Datasets Part 1 + Part 2

amcisaac:
I have combined the NAGL part 1 and part 2 datasets, and filtered out problematic records (crashes, geometric rearrangements, and stereochemistry). We may want to filter for Conformer RMSD, but I thought we would save that for the single point dataset creation if we do want to do it, let me know what you think.

To combine the two optimization datasets and have QCA recognize they had already been computed, I had to add the initial molecule to the dataset, not the optimized one. It _should_ then recognize the hash and skip the computation, and I confirmed that the hashes were the same between the separate and combined datasets.

Also, I called it `v4.0`, but I'm not sure if it should be called `v4.1` since it has been filtered? It would be the first version of the combined dataset, which is why I picked `v4.0`, but I'm not sure how this fits into the naming convention. Happy to change it.

It was non-trivial to combine these two datasets into one, partly due to it being an optimization dataset, and partly because of the large size of the dataset, so I am even more in favor of @ntBre's suggestion of tagging individual records than I was before.

## New Submission Checklist

- [x] Ready to submit!


ntBre (approved):
Thanks for doing this, Lexie! I was working on something similar last week but by combining the `dataset.json.bz2` files directly. This is a little more complicated but has the obvious benefit of being able to filter the records, so I think this looks good. Also glad to see the hashes check! Hopefully QCArchive will detect this as complete right after submission.

PR #418, opened 2024-12-13
Release sage 2.0.0

jaclark5:
## New Submission Checklist

- [NA] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [ ] Ready to submit!


submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -0,0 +1,147 @@
...
+for records in records_by_cmiles.values():
+    base_record, base_molecule = records[0]
+    base_molecule._conformers = [m.conformers[0] for _, m in records]
+
+    dataset1.add_molecule(
+        index=base_molecule.to_smiles(
+            isomeric=True, explicit_hydrogens=False, mapped=False
+        ),
+        molecule=None,
+        initial_molecules=[rec.initial_molecule for rec, _ in records],
```
ntBre, line 68: (not a review) Just for your future reference, this is the key difference forcing the use of the existing (`qcelemental`) molecules. The default usage here is to provide an `openff.toolkit.Molecule` to the `molecule` kwarg, which does the round-trip through the toolkit  Lexie mentioned in her comments. Providing `molecule=None` and `initial_molecules` (which are `qcelemental.models.Molecule`s) instead bypasses the round trip.

ntBre (commented):
This looks great overall! I think all of my comments are just subjectively what I think we "usually" do, and not necessarily requirements. I know you said on Slack this isn't ready for approval, but just to make a note, there are some missing files too:
* The input conda environment
* The final, fully-resolved conda environment (from `conda env export`)
* The output of the Python script

This last one is bordering on a "nice to have" rather than a strict requirement. I've definitely made some submissions without it, but it helps to fill the gaps between the Jupyter notebook submissions and a plain script (which I prefer for reviewing). I think `*.log` files are ignored in this repo, so make sure you save it as `output.txt` or something instead.

This one is covered by the checklist, but you'll also want to update the table in the main README.

Again, it looks good overall, so I'd usually drop a preemptive `Approve` here, but given the circumstances, I think I'll leave explicit approval to Lily.

Btw, nice job with the git-lfs stuff. I definitely messed that up on my first dataset.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -1,18 +1,19 @@
...
 from openff.toolkit.utils import OpenEyeToolkitWrapper, ToolkitRegistry
 from openff.units import unit
-import logging
-logging.getLogger("openff").setLevel(logging.ERROR)
 
-filtered_and_combined = OptimizationResultCollection.parse_file(
-    "/Users/jennifer.clark/bin/openff-sage/data-set-curation/quantum-chemical/data-sets/1-2-0-opt-set-v3.json",
-#    "/Users/jennifer.clark/bin/openff-sage/data-set-curation/quantum-chemical/data-sets/1-2-0-td-set.json",
-)
+file = requests.get("https://raw.githubusercontent.com/openforcefield/openff-sage/refs/heads/main/data-set-curation/quantum-chemical/data-sets/1-2-0-opt-set-v3.json")
```
ntBre, line 15: If you want to super, super pedantic, you can use the GitHub permalink: https://raw.githubusercontent.com/openforcefield/openff-sage/37a36e7eeaf6cdca795847089a288bdff168c08a/data-set-curation/quantum-chemical/data-sets/1-2-0-opt-set-v3.json. It's kinda annoying to get this (click the `...` in the upper right, copy permalink, then click `raw` for the raw file at the permalink), but it should always point to this exact file version instead of whatever `main` happens to be. This is especially unlikely to matter here because this repo is basically an archive anyway, though.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/README.md
```diff
@@ -0,0 +1,50 @@
...
+### Description
+
+A quantum chemical (QC) dataset curated to train [OpenFF 2.0.0 Sage](https://github.com/openforcefield/openff-sage) forcefield, with reparametrized Lennard-Jones (LJ) and valence parameters, the latter relevent to this dataset. This QC dataset with the OpenFF default level of theory, B3LYP-D3BJ/DZVP, is used to benchmark Sage geometries and energetics. These optimized conformer geometries where used in conjunction with the QC dataset used to train one dimensional torsional profiles. This Generation 2 dataset increases chemical diversity when compared to Generation 1, which are of value to our industry partners. Large molecules (>20 heavy atoms) were also included, including more flexible molecules and a greater degree of conformational variation which provide intramolecular interactions.
+
+### General Information
+
+- Date: 2024 12 12
+- Class: OpenFF Optimization Dataset
+- Purpose: B3LYP-D3BJ/DZVP conformers applicable to drug-like molecules for OpenFF 2.0.0 Sage
+- Collection: OptimizationDataset
```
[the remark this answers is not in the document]
jaclark5: It came with the template README.md file when I followed the User Quickstart instructions. Is this not needed now?

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/README.md
```diff
@@ -0,0 +1,50 @@
...
+- Mean molecular weight: 261.38
+- Max molecular weight: 544.64
+- Set of charges: -2.0 -1.0 0.0 1.0
+- Dataset Submitter: Jennifer A. Clark
+- Dataset Curator: Simon Boothroyd
+- Dataset Generator: Hyesu Jang
+
+### QCSubmit generation pipeline
+
+- `generate-combined-dataset.ipynb`: A notebook which shows how the dataset was prepared from the input files.
```
lilyminium, line 27:
```suggestion
- `generate-combined-dataset.py`: A script which shows how the dataset was prepared from the input files.
```
I'd also add `output.txt` since that's the output of this script.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -0,0 +1,149 @@
...
+print('Finished converting to records',flush = True)
+
+# Need to input initial molecules so that QCA will recognize the inputs as the same.
+initial_mols = [rec[0].initial_molecule for rec in rec_and_mol]
+
+dataset_factory1 = OptimizationDatasetFactory()
+provenance1 = dataset_factory1.provenance(ToolkitRegistry([RDKitToolkitWrapper]))
+
+dataset1 = OptimizationDataset(
+    dataset_name="OpenFF Sage 2.0.0 Training Optimization v1.0",
```
lilyminium, line 29: As above -- could the name of this dataset please match the directory name?

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -0,0 +1,149 @@
...
+rec_and_mol = filtered_and_combined.to_records()
+
+print('Number of results: ',filtered_and_combined.n_results,flush=True)
+print('Finished converting to records',flush = True)
+
+# Need to input initial molecules so that QCA will recognize the inputs as the same.
+initial_mols = [rec[0].initial_molecule for rec in rec_and_mol]
+
+dataset_factory1 = OptimizationDatasetFactory()
+provenance1 = dataset_factory1.provenance(ToolkitRegistry([RDKitToolkitWrapper]))
```
lilyminium, line 26: Just looking at your environment, it looks like OpenEye is installed. Do you have a license? If so, it takes precedence over RDKit so probably best to add that here too.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -0,0 +1,149 @@
...
+        "https://github.com/openforcefield/qca-dataset-submission/tree/master/"
+        "submissions/"
+        "2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0"
+    )
+
+
+# Have to add records this way to avoid a round trip through the toolkit.
+records_by_cmiles= {}
+for record, molecule in rec_and_mol:
+    cmiles = molecule.to_smiles(isomeric=True, explicit_hydrogens=True, mapped=True)
```
lilyminium, line 54: A bit nitpicky, but since this might establish a protocol for future tasks -- there's no guarantee that running a `to_smiles` will give you the same cmiles for the same molecule across different OpenEye/RDKit versions, or especially if you have one toolkit installed but another was used to generate the source datasets. It would be slightly more robust to use the exact cmiles in the dataset result. The easiest way to do this is probably by inspecting `filtered_and_combined.entries`, which is a dictionary where the key is the portal address and the value is a list of `OptimizationResult`s. Each `OptimizationResult` has an `id` and `cmiles` attribute that should correspond to the `record.id` of each record in `rec_and_mol`. Otherwise there may be other ways similar to what QCSubmit does when it downloads datasets (https://github.com/openforcefield/openff-qcsubmit/blob/d4e6b6986a58f5cf0184ba14dc4f7419e9978b67/openff/qcsubmit/results/results.py#L307-L353), but I haven't looked into those.

I'm not sure this actually matters very much; at the end of the day an OptimizationResultCollection or OptimizationDataset is a flat list of single-conformer results, and functionality like our ConformerRMSDFilter checks whether each record is a conformer of an existing molecule on-the-fly. However, it would be more ideal to keep the cmiles consistent between datasets just in case it does in the future.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
lilyminium: Thanks Brent, helpful to know!

lilyminium (changes requested):
This mostly looks great -- thanks @jaclark5 for putting this together and @ntBre for the thorough review! Mine largely echoes Brent's comments, but I'm also not sure that the CMILES being used to index molecules between datasets will remain the same between software versions; it would be more robust to just use the CMILES in the original dataset, as commented.

Outside the scope of this PR but just a thought -- we should also consider having a separate table for force field datasets in the top-level README, otherwise it's a slight hassle to find.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
jaclark5: Oh thanks for this! I did notice that the CMILES changes depending on whether I use RDKit or OpenEye.

ntBre: Good catch Lily! Just in Jen's defense, I believe this code came straight from Lexie's NAGL2 combined dataset script. So I've also reviewed and approved it twice now, missing this subtlety. Also this is how `OptimizationResultCollection.create_basic_dataset` is implemented (except there `explicit_hydrogens=False, mapped=False` are passed), so Lexie probably got the code from there anyway.

This section of qcsubmit contains all of the locations you can check for a CMILES in an `OptimizationResultCollection`: https://github.com/openforcefield/openff-qcsubmit/blob/d4e6b6986a58f5cf0184ba14dc4f7419e9978b67/openff/qcsubmit/results/results.py#L483-L496. You can also be a bit clever with bool chaining:

```python
 def get_cmiles(entry):
     mol = entry.initial_molecule
     return (
         mol.identifiers.canonical_isomeric_explicit_hydrogen_mapped_smiles
         or mol.extras.get("canonical_isomeric_explicit_hydrogen_mapped_smiles")
         or entry.attributes.get(
             "canonical_isomeric_explicit_hydrogen_mapped_smiles"
         )
     )
```

But `entry` here has to be fetched from qcportal, so I agree that pulling the CMILES from the `OptimizationResultCollection` itself would be easier.

As a final observation, I think this line is not actually the problematic use of `to_smiles` in the script because this `cmiles` is only used to key a dict. `to_smiles` is called again on line 65, where it's actually stored in the dataset and would cause the kind of problem Lily is pointing out.

lilyminium: Thanks, good catches Brent, yes line 65 is definitely where it matters!

You're more familiar with the code here than me -- if you think this will be problematic for `create_basic_dataset`, could you please raise an issue in qcsubmit?

@jaclark5 please ping me when you'd like a re-review -- I have notifications turned off for PR pushes so won't see any changes made.

ntBre: Just opened openforcefield/openff-qcsubmit#310 referencing this.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
lilyminium: Just following up on this, it's less important if CMILES isn't being generated, but if RDKit is there OpenEye should be considered too.

lilyminium (commented):
LGTM @jaclark5, just one final question on OpenEye!

I'll also add a `tracking` label to this -- in QCA-DS a lot of action happens in the CI, which monitors pull requests with particular labels to trigger workflows. The `tracking` label means a PR gets tracked on our [project board](https://github.com/orgs/openforcefield/projects/2?query=sort%3Aupdated-desc+is%3Aopen), which in turn has associated CI workflows to manage dataset submission and error cycling. In general we try to add the `tracking` label and any `compute-` labels before a PR gets merged, although we don't expect this one to require any additional computation.

submissions/2024-12-12-OpenFF-Sage-2.0.0-Training-Optimization-Dataset-v1.0/generate-combined-dataset.py
```diff
@@ -12,156 +14,179 @@
 from openff.toolkit.utils import OpenEyeToolkitWrapper, ToolkitRegistry
 from openff.units import unit
 
-def pull_record_id_cmiles(Opt):
 
+def pull_record_id_cmiles(Opt: Type[OptimizationResultCollection]):
```
lilyminium, line 18: (non-blocking since the type signature won't affect performance) -- I hadn't actually seen typing.Type used before so had to look it up. It looks like it's an alias of `type`? If so I think `Opt: OptimizationResultCollection` -- this tells the typer that `Opt` should be an object of OptimizationResultCollection, whereas `type` would refer to a class itself.

lilyminium: If black did this automatically, it might be because of the capitalization of Opt confusing it.

lilyminium (approved):
LGTM -- please merge when you're ready!

PR #425, opened 2025-01-21
Tag debug

jaclark5:
The automated molecular weight size binning and tagging had issues derived from record ids that were numpy.int64 type instead of python int type. This is now resolved.


j-wags (approved):
Looks good, thanks @jaclark5!

PR #430, opened 2025-02-26
Updating validation environment to update units package

### LINKED ISSUE #429: Validation not passing due to using `openff-units` 0.1.8 and `pint 0.24`
I got this error in my validation:

```Traceback (most recent call last):
  File "/home/runner/work/qca-dataset-submission/qca-dataset-submission/./management/validation.py", line 14, in <module>
    from openff.qcsubmit.datasets import (BasicDataset, OptimizationDataset,
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/qcsubmit/datasets/__init__.py", line 1, in <module>
    from openff.qcsubmit.datasets.dataset_utils import (
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/qcsubmit/datasets/dataset_utils.py", line 7, in <module>
    from openff.qcsubmit.datasets.datasets import (
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/qcsubmit/datasets/datasets.py", line 20, in <module>
    from openff.toolkit import topology as off
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/toolkit/__init__.py", line 91, in __getattr__
    return importlib.import_module(lazy_mod)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/validation/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/toolkit/topology/__init__.py", line 1, in <module>
    from openff.toolkit.topology.molecule import (
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/toolkit/topology/molecule.py", line 49, in <module>
    from openff.units import Unit
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/units/__init__.py", line 4, in <module>
    from openff.units.units import DEFAULT_UNIT_REGISTRY, Measurement, Quantity, Unit
  File "/home/runner/micromamba/envs/validation/lib/python3.11/site-packages/openff/units/units.py", line 11, in <module>
    from pint.measurement import _Measurement
ModuleNotFoundError: No module named 'pint.measurement'
```

It looks like it's being caused by the combination of `openff-units 0.1.8` and `pint 0.24`. 

@mattwthompson suggested pinning to `openff-units >=0.2.0`, I can open a PR to patch it.

amcisaac:
Pinned the version of `openff-units` in the validation environment to avoid import error described in #429 . Fixes #429 for me locally, but I can't run the full validation locally due to missing a github token.

### DIFF
--- a/devtools/conda-envs/error-cycle.yaml
+++ b/devtools/conda-envs/error-cycle.yaml
@@ -6,9 +6,10 @@ channels:
 dependencies:
   - python
   - qcportal >=0.49
-  - openff-qcsubmit >=0.53.0
+  - openff-qcsubmit >=0.55.0
   - openff-toolkit-base >=0.12
   - openff-forcefields >=2.0.0
+  - openff-units >= 0.2.0
   - PyGithub
   - pandas
   - tabulate
--- a/devtools/conda-envs/queued-submit.yaml
+++ b/devtools/conda-envs/queued-submit.yaml
@@ -6,9 +6,10 @@ channels:
 dependencies:
   - python
   - qcportal >=0.49
-  - openff-qcsubmit >=0.53.0
+  - openff-qcsubmit >=0.55.0
   - openff-toolkit-base >=0.12
   - openff-forcefields >=2.0.0
+  - openff-units >= 0.2.0
   - PyGithub
   - pandas
   - tabulate
--- a/devtools/conda-envs/validation.yaml
+++ b/devtools/conda-envs/validation.yaml
@@ -6,7 +6,7 @@ channels:
 dependencies:
   - python
   - qcportal >=0.49
-  - openff-qcsubmit >=0.53.0
+  - openff-qcsubmit >=0.55.0
   - openff-toolkit-base >=0.12
   - openff-forcefields >=2.0.0
   - openff-units >= 0.2.0
--- a/devtools/conda-envs/validation.yaml
+++ b/devtools/conda-envs/validation.yaml
@@ -9,6 +9,7 @@ dependencies:
   - openff-qcsubmit >=0.53.0
   - openff-toolkit-base >=0.12
   - openff-forcefields >=2.0.0
+  - openff-units >= 0.2.0
   - basis_set_exchange
   - pydantic
   - typing-extensions


amcisaac:
Just requesting reviews from the folks I know are working on this repo--Let me know if there's a better person to request!

amcisaac:
Sorry for all the comments--just noting that this environment solves (for me locally) to use QCSubmit 0.54, which doesn't allow the custom convergence criteria. I could also pin to QCSubmit >= 0.55, but not sure if that's desired or if that requires any other changes/testing

j-wags:
I think @jaclark5 should be good to review this (though feel free to tag me in if there are questions, I'm just booked up with other stuff for the next few days). 

On the level of caution to use here - While pushing to prod is usually harrowing, this is a weird repo that doesn't follow normal release/versioning rules, and in practice issues with bad envs/deps will very likely cause an error to be raised instead of letting a mangled dataset to be submitted. There also aren't tests that will tell us when things jump the rails, so manual checking is the best we can do here. 

I think this PR should up-pin qcsubmit to >=0.55. On a good day that wouldn't be needed but our ecosystem is in a weird state right now. So I'd recommend adding that and any other up-pins that are necessary to get a sane env to solve. 

I'd take some time to think about whether other envs in the same folder will need to be up-pinned as well.

amcisaac:
Okay, I'll pin QCSubmit too for validation, since it's not solving to that version for some reason. Definitely the dataset submission part of `lifecycle.py` needs to use QCSubmit > 0.55 for my dataset, and I assume anything that imports from `openff` will call `units` somewhere, so that should probably be pinned-- It sounds like pinning `units` is relatively fool proof since the old `0.1.8` isn't meant to be used anymore, but QCSubmit might be iffier

jaclark5:
Based on the original error, the issue from `openff-units` is thrown via `open-toolkit`, rather than updating the constraint for the benefit of the dependency, I like the idea of pinning `openff-qcsubmit >= 0.55` instead. Since it has an internal constraint `openff-units ~=0.2.1` it should also solve the problem and will ensure that each step has the needed changes with the `maxiter` keyword.

If we are pinning `openff-qcsubmit >= 0.55` in validation.yaml, I think we should also pin it in queued-submit.yaml and error-cylce.yaml.

amcisaac:
Hm, the validation is using units 0.1.8 despite also using qcsubmit 0.55, and that's what it solves to on my local machine as well if I don't pin units...maybe there's an issue with the QCSubmit dependencies? https://github.com/openforcefield/qca-dataset-submission/actions/runs/13556344431/job/37943912062?pr=428

jaclark5:
Ok this isn't worth holding it up, let's do open-units and qcsubmit.

jaclark5 (approved):

PR #436, opened 2025-03-13
Update ci

jaclark5:
Update error cycling to accept scaffold.json files that bypass QCSubmit. 
Also resolve un-updated pr label lists with preferred GitHub method.

### DIFF
--- a/management/lifecycle.py
+++ b/management/lifecycle.py
@@ -153,7 +153,7 @@ def update_compute_tags(client, dataset, specification_names, new_tag, include_c
         set_mw_compute_tags(client, dataset, new_tag, include_complete=include_complete)
 
 def _get_labels(pr):
-    return [ label.name for label in pr.get_labels() ]
+    return list(set(label.name for label in pr.get_labels()))
 
 class Submission:
     """A submission, corresponding to a single PR, possibly multiple datasets.
--- a/management/lifecycle.py
+++ b/management/lifecycle.py
@@ -1104,7 +1104,7 @@ def main():
         # take highest one and set priority downstream
         # if no priority label(s), DO NOT set priority at all for this PR
         if args.set_priority:
-            labels =  _get_labels(pr)
+            labels =  set(_get_labels(pr))
             priorities = set(PRIORITIES.keys()) & labels
 
             if not priorities:
--- a/management/lifecycle.py
+++ b/management/lifecycle.py
@@ -22,6 +22,7 @@ QCFRACTAL_URL = "https://api.qcarchive.molssi.org:443/"
 
 REPO_NAME = "openforcefield/qca-dataset-submission"
 DATASET_GLOB = "dataset*.json*"
+SCAFFOLD_GLOB = "scaffold*.json*"
 COMPUTE_GLOB = "compute*.json*"
 
 PRIORITIES = {'priority-low': 0, 'priority-normal': 1, 'priority-high': 2}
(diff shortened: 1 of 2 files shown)


lilyminium (approved):
LGTM, nothing jumps out as obviously possibly problematic; there's no real test suite for this repo and it seems easier to run through CI to test, so happy with merging as is!

management/lifecycle.py
```diff
@@ -1096,7 +1104,7 @@ def main():
         # take highest one and set priority downstream
         # if no priority label(s), DO NOT set priority at all for this PR
         if args.set_priority:
-            labels =  set(map(lambda x: x.name, pr.labels))
+            labels =  _get_labels(pr)
```
j-wags, line 1107: Since this gets &'ed in the next line, we need to ensure that this is a set
```suggestion
            labels =  set(_get_labels(pr))
```

management/lifecycle.py
```diff
@@ -205,8 +208,9 @@ def __init__(self, pr, ghapi, repo=None, priority=1, computetag='openff'):
     def _gather_datasets(self):
         files = self.pr.get_files()
         datasets = list(filter(
-            lambda x: glob.fnmatch.fnmatch(os.path.basename(x), DATASET_GLOB),
-            map(lambda x: x.filename, files)))
+            lambda x: any(glob.fnmatch.fnmatch(os.path.basename(x), match) for match in [DATASET_GLOB, SCAFFOLD_GLOB]),
+            map(lambda x: x.filename, files)
+        ))
```
j-wags, line 213: I think this might need to be a list concatenation to avoid messing up the iteration depth (but am not positive, hard to follow everything going on in this command) ... Also this is turning into a pretty gnarly list comprehension that'll be hard to debug in the future, so I'd advise refactoring it into several separate commands. 

```suggestion
            lambda x: any(glob.fnmatch.fnmatch(os.path.basename(x), match) for match in [DATASET_GLOB + SCAFFOLD_GLOB]),
            map(lambda x: x.filename, files)
        ))
```

j-wags (approved):
Two proposed changes but otherwise this should be good to go

management/lifecycle.py
jaclark5: I don't think this is right since the variables are strings, this would combine the following two strings defined at the top of the file:
```
DATASET_GLOB = "dataset*.json*"
SCAFFOLD_GLOB = "scaffold*.json*"
```

management/lifecycle.py
jaclark5: Instead of this change I altered `_get_labels` to return `list(set(label.name for label in pr.get_labels()))`

management/lifecycle.py
jaclark5: From the qca-dataset-submission management directory:
```
>>> any(glob.fnmatch.fnmatch(os.path.basename("validation.py"), match) for match in ["validation*", "lifecycle*"])
True
>>> any(glob.fnmatch.fnmatch(os.path.basename("lifecycle.py"), match) for match in ["validation*", "lifecycle*"])
True
>>> any(glob.fnmatch.fnmatch(os.path.basename("project*"), match) for match in ["validation*", "lifecycle*"])
False
```

management/lifecycle.py
j-wags: I don't think that'll help - the `&` operator isn't defined for lists

management/lifecycle.py
j-wags: Ah, good catch. Not needed, then.

management/lifecycle.py
j-wags: You may need to look at this in the file diff tab to see what I mean, since the line with the `&` is after the line where the change is proposed

PR #432, opened 2025-03-06
Add PDB 4-mers dataset

ajfriedman22:
## New Submission Checklist

- [x] Created a new folder in the submissions directory containing the dataset
- [x] Added `README.md` describing the dataset see [here](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2020-03-26-OpenFF-Gen-2-Torsion-Set-6-supplemental-2) for examples
- [x] All files used to produce the dataset are included with a description
- [x] Dataset follows the QCSubmit schema defined for [Datasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L340), [OptimizationDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1062) and [TorsionDriveDatasets](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L1225)
- [x] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [x] A PDF depicting the molecules is attached, in the case of torsiondrives this should include the highlighting of the central bond, this can be done automatically using [qcsubmit](https://github.com/openforcefield/qcsubmit/blob/56680a7d3298b5d8962edcb840b0fdb34558c053/qcsubmit/datasets.py#L854). 
- [x] QCSubmit validation passed
- [x] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!

PR #443, opened 2025-03-24
Add constraints to ferrocene complexes in pr435

jaclark5:
The xtb package fails to capture haptic bonds such as in ferrocene. This PR adds constraints to the ferrocene complexes to freeze the bond length between Fe and each of the carbon atoms in the adjacent rings.


lilyminium (changes requested):
Thanks, just a few questions:

* Checking #435, this dataset has already been submitted and started compute without constraints -- is that correct? If so, could you please instead increment the version to 0.1 and submit a new PR detailing what was changed, instead of overwriting the old one?
* I'm having a very hard time finding exactly where MolSSI documents how to add constraints -- would you have a link? The closest I can find is in the [qcsubmit conversion](https://github.com/openforcefield/openff-qcsubmit/blob/711a17cc8e53873d1fc70757df9f60d4774f38ec/openff/qcsubmit/datasets/datasets.py#L1040-L1071) where the `additional_keywords` is of a form `{"constraints": ...something...}`, whereas the notebook here seems to drop the "constraints" key.

submissions/2025-03-12-PDB-CCD-Chemical-Space-A-Optimization-Dataset-v0.0/generate-dataset.ipynb
```diff
@@ -409,7 +537,7 @@
     "for label, mol_dict in output.items():\n",
     "    \n",
     "    rdconf = mol_dict['rdmol'].GetConformer()\n",
-    "    conformer = [list(x) for x in rdconf.GetPositions()]\n",
+    "    conformer = [list(x) for x in rdconf.GetPositions() / constants.bohr2angstroms] # Convert from nm to Bohr (a0)\n",
```
lilyminium, line 540: Just checking from the comment -- is RDKit in angstrom or nm?

jaclark5:
> * I'm having a very hard time finding exactly where MolSSI documents how to add constraints -- would you have a link? The closest I can find is in the [qcsubmit conversion](https://github.com/openforcefield/openff-qcsubmit/blob/711a17cc8e53873d1fc70757df9f60d4774f38ec/openff/qcsubmit/datasets/datasets.py#L1040-L1071) where the `additional_keywords` is of a form `{"constraints": ...something...}`, whereas the notebook here seems to drop the "constraints" key.

Yeah it doesn't seem to be documented. I ended up deducing how since constraints are implemented in QCSubmit, I ran #432 (Peptide 4-mer) data through `scaffold.py` to find that the constraints live it `additional_keywords` for an entry. This is because a constraint is actually a keyword that "should" live in the specification, but since it would only apply to one entry and we don't want a new specification for each entry, it "lives" in the entry under `additional_keywords`. At least that's my theory. You can see that `additional_keywords` is an input for OptimizationDatasetEntry [in the source code](https://github.com/MolSSI/QCFractal/blob/f1466641cd59332711a5a4890d786708ea7cc1de/qcportal/qcportal/optimization/dataset_models.py#L21).

jaclark5:
> * Checking [TM PDB CCD Contribution #435](https://github.com/openforcefield/qca-dataset-submission/pull/435), this dataset has already been submitted and started compute without constraints -- is that correct? If so, could you please instead increment the version to 0.1 and submit a new PR detailing what was changed, instead of overwriting the old one?

This PR also scales the coordinates from Angstrom to Bohr, so I plan to delete all of the old records and replace with this scaffold.json.

jaclark5:
From a scaffold of #432:
```
    "entries": {
        "AAAA-0": {
            "name": "AAAA-0",
            "initial_molecule": {
                 ...
             },
            "additional_keywords": {
                "constraints": {
                    "freeze": [
                        {
                            "type": "dihedral",
                            "indices": [
                                1,
                                3,
                                4,
                                5
                            ]
                        },
                       ...
                  ]}},

jaclark5:
> Ok, thanks! It might be reaching out to Ben (or cc @j-wags if he knows off the top of his head) to confirm. If I'm reading correctly the source code I linked does create a `OptimizationDatasetEntry.additional_keywords` with a dictionary where constraints aren't top level, but instead under the "constraints" key. So (without testing this, just reading code) I'm interpreting QCSubmit as generating a `{"constraints": {"freeze": ...}}` dictionary, whereas your code directly generates a `{"freeze": ...}` dictionary.

Good catch, I updated that in the notebook

submissions/2025-03-12-PDB-CCD-Chemical-Space-A-Optimization-Dataset-v0.0/generate-dataset.ipynb
lilyminium: Thanks!

jaclark5:
> Largely LGTM! Just one final question -- I can see the final commit updated the scaffold and PDF, so I _assume_ the notebook was re-executed, but there's an error in the actual committed notebook -- would you mind re-executing and just checking? Thanks!

There isn't an error in the notebook, but it seems that GitHub is failing to render the 3D representation of the molecule. I have all green checkmarks locally with nothing to commit.

lilyminium:
Thanks for re-uploading -- everything looks great!

PR #456, opened 2025-05-05
Update Docker image to allow for basis set to be updated

jaclark5:
Update the docker file to allow for aug-cc-pvtz-dk basis set to be updated.


devtools/prod-envs/qcarchive-worker-openff-psi4-basis-set-update.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - dftd3-python
+  - gcp-correction
+  - scipy
+  - dftd4-python
+    
+  # procedures
+  - geometric
+
+  - pip:
+    - basis_set_exchange
```
mattwthompson, line 24: There's a conda package for this (of the same name)

j-wags: (not blocking) The conda package seems to be kept up to date, but it's not referenced in the [docs](https://molssi-bse.github.io/basis_set_exchange/usage.html) or [repo README](https://github.com/MolSSI-BSE/basis_set_exchange), both of which point to the package on pypi. It's likely fine, just something to keep an eye on if it stops being updated.

devtools/prod-envs/qcarchive-worker-openff-psi4-basis-set-update.yaml
```diff
@@ -0,0 +1,23 @@
+name: qcarchive-worker-openff-psi4
+channels:
+  - conda-forge/label/libint_dev
+  - conda-forge
+dependencies:
+  - python =3.10
```
j-wags, line 6: (not blocking) We'll want to stop pinning to python 3.10 for our images in the next few months, since it's falling out of the SPEC0 range. I'll open an issue to track this.

j-wags (approved):
Two non-blocking comments, so this should be good to merge! LGTM, thanks for your persistence on this, @jaclark5

devtools/prod-envs/qcarchive-worker-openff-psi4-basis-set-update.yaml
jaclark5: I've been using 3.11 locally so I have confidence that that the runners will be fine with that. I'll update it to that.

PR #460, opened 2025-05-22
add updated compute tag kwarg

### LINKED ISSUE #459: modify_records broken
new_tag is no longer an accepted kwarg in BaseDataset.modify_records, it's been replaced with new_compute_tag. The new QCPortal client.modify_records function looks for the old deprecated argument but not the BaseDataset function.

lilyminium:
Fixes #459 and also updates to compute_priority instead of priority


lilyminium:
Self-merging so I can retag my PR

PR #475, opened 2025-06-26
Industry benchmark v1.2

jaclark5:
## New Submission Checklist

- [X] Dataset filename matches pattern `dataset*.json`; may feature a compression extension, such as `.bz2`
- [ ] Validation passed
- [X] Made a new dataset entry in the mapping table in repository `README.md`
- [ ] Ready to submit!


lilyminium (changes requested):
Thanks for starting this, Jen. The numbers here look a bit fishy to me -- I'd expect ~70k conformers to start with (even after applying all the post-submission filters) and the outliers txt file only has 29 IDs, so I'm surprised you wind up with 50k conformers. Did everything download? Could you maybe double check numbers in the dataset pre-filtering and post-filtering?

Speaking of post-submission filters, apologies for only remembering this now but we do run a set of post-submission filters on our datasets before we use them. Would it be possible to run those on the v1.1 dataset as well for v1.2 to get a closer representation of what we actually use? The list is [here](https://github.com/openforcefield/qca-dataset-submission/blob/master/STANDARDS.md#fitting-standards) and an example of applying filters with QCSubmit is [here](https://docs.openforcefield.org/projects/qcsubmit/en/stable/examples/retrieving-results.html#Filtering-result-collections). I also semi-frequently put together my own code for reasons of speed, e.g.:


```python
def single_process_optimization(item):
    record, molecule = item
    try:
        if not record.status.upper() == "COMPLETE":
            return
        if not ConnectivityFilter()._filter_function(None, None, molecule):
            return
        if not UnperceivableStereoFilter()._filter_function(None, None, molecule):
            return
    # can have some rdkit connection errors
    except:
        return
    return item


records_and_molecules = optimization_result_collection.to_records()
with multiprocessing.Pool(n_processes) as pool:
        results = list(
            tqdm.tqdm(
                pool.imap(single_process_optimization, records_and_molecules),
                total=len(items),
                desc="Processing records and molecules"
            )
        )
```

jaclark5:
This PR resulted in n_inserted=74585, n_existing=0 records. We expected n_existing=74585. The impact of this PR will be reversed and corrected with PR #483

PR #480, opened 2025-07-02
Add notebooks for each dataset type

jaclark5:
Update the archival materials to include notebooks for single datasets, either torsiondrive, optimization, or singlepoint.

- The optimization notebook is identical to the optimization portion of the original notebook, with the addition of a try/except statement to find the CMILES information.
- The torsiondrive notebook is identical to the optimization portion of the original notebook, with an alteration of the printed energies at the end.
- The singlepoint notebook is based on the optimization notebook with alterations in attribute names for functionality.


lilyminium (approved):
LGTM!

PR #483, opened 2025-07-03
Update industry benchmark dataset

jaclark5:
Even though the spec keywords were the same as the QCSubmit defaults, submitting Industry Benchmark v1.2 resulted in new records being made. This PR bypasses QCSubmit to create the new dataset, but still uses it to filter and write out a file containing record ids expects to be in the new dataset.


lilyminium (approved):
LGTM, thanks for fixing this! This might have to be the standard way to do this, which is unfortunate... perhaps the ideal solution is for QCSubmit to have this functionality too.

PR #484, opened 2025-07-10
Update archival notebooks to v1.1

jaclark5:
Loading all records into memory for some datasets crashes the kernel in the docker container. The new version of the notebook iterates through the records without loading them into memory.


j-wags (approved):
LGTM, thanks @jaclark5!

PR #539, opened 2026-09-15
Update container

jaclark5:
Update docker container with recent version of psi4 and qcelemental

This docker container / yaml combination was tested locally on my mac


jaclark5:
GitHub Action: https://github.com/openforcefield/qca-dataset-submission/actions/runs/35006429626

devtools/prod-envs/qcarchive-worker-openff-psi4.yaml
```diff
@@ -3,20 +3,21 @@ channels:
...
   - conda-forge
 dependencies:
-  - python =3.10
+  - python =3.13
   - pip
-  - qcfractalcompute =0.51
-  - qcengine =0.29.0
-  - qcelemental =0.27.1
+  - qcfractal
+  - qcfractalcompute>=0.7
```
j-wags: (not blocking) The most recent release was 0.70, and the [actual docker build](https://github.com/openforcefield/qca-dataset-submission/actions/runs/35006429626/job/104507217736#step:6:380) installs 0.70 even without this change, but it could be good to update it to avoid ambiguity



```suggestion
  - qcfractalcompute=0.70
```

j-wags (approved):
(not blocking) These yamls are a bit different than the normal ones we use for CI since they're aimed at creating specific envs and are kinda provenance-adjacent, so I'd recommend going back to using `=` instead of `>=` for most deps.

PR #540, opened 2026-09-15
Add cuest docker

jaclark5:
Add docker image for GPU workers


jaclark5:
GitHub Action https://github.com/openforcefield/qca-dataset-submission/actions/runs/35007338936

jaclark5:
@j-wags I was able to successfully create the docker image locally but I needed to make a new GitHub action to run it here, and that doesn't show up as an option until it's in the main branch. 

Do you have a recommendation on how I can test this before merger?

j-wags (approved):
Sorry for the delay @jaclark5, it's been a while since I worked with the docker build actions in this repo. 

I think I'm wrapping my head around it and you're right- the `prod-gpu-envs-docker.yml` file needs to be merged to `master` to be triggerable on the actions page. (alternatively all the `build-args` could be put into the vanilla github action `.yml` but that'd be messy, so I think your strategy of making the separate action is the best way forward). 

This is good to merge as-is. The current state doesn't have to be perfect since  you'll be able to start a second PR to iterate once the action `.yml` is on `master.

